PR #4, opened 2018-11-09
Add optional callback to Formbot#onChange

erikshestopal:
> ### Summary
I ran into a situation where I needed to call `onChange` for field B that depended on the `onChange` of field A (see example) below. I have also tried unsuccessfully to use the `onChange` prop on `Formbot`
to set a local state that I could then query. Adding this optional callback seemed to be the most straightforward way of doing it but other suggestions are always welcome.

```jsx
<Formbot>
  {({ values, onChange }) => (
    <Fragment>
      <Field>
        <Select
          value={values.status}
          onChange={value => {
            onChange('status', value);

            // `If we had selected a status and then decided to remove it,
            // values` won't update fast enough to reflect the new state and
            // `values.status` would still be set to the old status
            if (values.status.length === 0) {
              onChange('doctor', '');
            }

            // with a callback
            onChange('status', value, newValues => {
              if (newValues.status.length === 0) {
                // works now!
                onChange('doctor', '');
              }
            });
          }}
        />
      </Field>

      <Field>
        <Select value={values.doctor} onChange={value => onChange('doctor', value)} />
      </Field>
    </Fragment>
  )}
</Formbot>;

```

### DIFF
--- a/src/Form/Formbot.js
+++ b/src/Form/Formbot.js
@@ -161,7 +161,7 @@ export default class Formbot extends React.Component {
     });
   };
 
-  onChange = (field, value) => {
+  onChange = (field, value, callback) => {
     this.updateField(field, { validated: false }).then(() => {
       this.setState(
         {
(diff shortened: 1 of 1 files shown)


kylealwyn:
@erikshestopal Can you elaborate on the functionality you're trying to produce?

erikshestopal:
Kyle and I discussed this offline but there is a better way of handling this without needing the callback. 
🤦‍♂️ 😆 It was a good learning moment.

PR #6, opened 2018-11-13
Pass className prop to Input component's children

erikshestopal:
Necessary for getting `styled-components` to pass down class names to underlying components.

### DIFF
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -220,6 +220,7 @@ export default class Input extends Component {
     const inputProps = {
       ...rest,
       id,
+      className,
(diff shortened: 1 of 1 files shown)


src/Form/Input.js
```diff
@@ -220,6 +220,7 @@ export default class Input extends Component {
     const inputProps = {
       ...rest,
       id,
+      className,
```
kylealwyn: Can just delete className from L199 so it'll fall into `...rest`

erikshestopal:
@kylealwyn can you release this please?

PR #8, opened 2018-11-29
fix(Formbot): Update minLength and maxLength error messages

choochootrain:
discovered via sappira-inc/web-app#116


choochootrain:
lol @ https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident
i updated the deps that were pulling this in

jlemm45:
lgtm after checks fixed

PR #9, opened 2018-12-01
fix(RadioGroup): don't loop state updates

choochootrain:
looks like this was the issue here: https://app.asana.com/0/821652619613450/929543341861746/f

`componentDidUpdate` gets called after `handleChange` on user input - this was calling `handleChange` again with the old value (because `this.props.value` isn't updated until `Formbot.setState` is called and propagated back down)

### DIFF
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -76,20 +76,22 @@ export default class Checkbox extends React.Component {
     disabled: false,
   };
 
+  static getDerivedStateFromProps(props, state) {
+    if (props.value && props.value !== state.value) {
+      return {
+        value: props.value,
+      };
+    }
+
+    return null;
+  }
+
   state = {
     value: this.props.value,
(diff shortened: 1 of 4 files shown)


kylealwyn:
Discussed irl but documenting to also update https://github.com/sappira-inc/molekule/blob/master/src/Form/Checkbox.js#L88 to use `getDerivedStateFromProps`

kylealwyn:
@choochootrain I'm also maybe a potential with the current logic. If a prop other than `value` changes, it'll always update the internal state value to the external value, which in this case would be undefined.

Propose logic to look like: 

```js
if (props.value && props.value !== state.value) {
  return {
    ...state, // cant remember if this shallow merges?
    value: props.value,
  };
}

return null;
```

choochootrain:
:+1: good catch, i'll update. the derived state merges like `setState` which is nice

PR #10, opened 2018-12-05
Don't use dangerous HTML in Linkify

erikshestopal:
> ### Summary
Fixed an issue with the `Linkify` component allowing execution of arbitrary HTML/JS that could be used for XSS attacks. 

Props to @kylealwyn for the actual implementation.

### DIFF
--- a/src/Linkify/Linkify.js
+++ b/src/Linkify/Linkify.js
@@ -2,45 +2,48 @@ import React from 'react';
 
 const URL_REGEX = /(\b(https?|ftp|file):\/\/[-A-Z0-9+&@#/%?=~_|!:,.;]*[-A-Z0-9+&@#/%=~_|])/gi;
 
-const detectLinks = ({ children, linkStyle = {} }) =>
+const Link = ({ href, linkStyle }) => (
+  <a
+    target="_blank"
+    rel="noopener noreferrer"
+    style={{
+      color: 'inherit',
+      textDecoration: 'underline',
+      ...linkStyle,
+    }}
+    href={href}>
+    {href}
+  </a>
+);
+
+const linkify = ({ children, linkStyle = {} }) =>
   React.Children.map(children, child => {
     if (!child) return null;
 
     if (typeof child === 'string') {
-      const words = child.split(' ');
-      return words.map((word, index) => {
-        const isLastWord = words.length - 1 === index;
+      const matches = child.match(URL_REGEX);
 
-        if (URL_REGEX.test(word)) {
-          return (
-            <>
-              <a
-                key={word}
-                target="_blank"
-                rel="noopener noreferrer"
-                style={{
-                  color: 'inherit',
-                  textDecoration: 'underline',
-                  ...linkStyle,
-                }}
-                href={word}>
-                {word}
-              </a>
+      if (!matches) {
+        return child;
+      }
 
-              {isLastWord ? '' : ' '}
-            </>
-          );
-        }
+      // Clone the string for manipulation
+      let string = `${child}`;
 
-        return isLastWord ? word : `${word} `;
-      });
+      // Run through the url matches to create our array of string parts and Link components
+      const stringParts = matches.reduce((parts, url) => {
+        const urlStartIndex = string.indexOf(url);
+        const partBeforeUrl = string.substring(0, urlStartIndex);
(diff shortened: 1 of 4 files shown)


src/Linkify/Linkify.js
```diff
@@ -8,9 +8,21 @@ const Container = styled.span`
...
+const REGEX = /(\b(https?|ftp|file):\/\/[-A-Z0-9+&@#/%?=~_|!:,.;]*[-A-Z0-9+&@#/%=~_|])/gi;
+
+const parse = string =>
+  escapeHtmlChars(string).replace(
+    REGEX,
+    url => `<a target="_blank" rel="noopener noreferrer" href="${url}">${url}</a>`
+  );
 
 const Linkify = props => (
   <Container style={props.style || {}} dangerouslySetInnerHTML={{ __html: parse(props.children) }} />
```
choochootrain: i think you can do this without `dangerouslySetInnerHTML` if you iterate through regex matches and return a `[React.Component]`. see https://github.com/tasti/react-linkify/blob/master/src/components/Linkify.jsx#L75 (could also use that library directly, up to you)

erikshestopal: While I think using a library like this would be ideal, our use case for this component usually resolves parsing strings as opposed to React elements. So, for example, we currently use the component like this:
```jsx
<Linkify>
  {message.body} //=> string
</Linkify>
```
as opposed to like this. 
```jsx
<Linkify>
  <span>Hello this a tag</span>
  <span>https://link.com</span>
</Linkify>
```
While I definitely do agree about with the general notion of making code as general and extensible as possible, I think in this case I would hold off on addressing edge cases until and if we get to them. 

What do you think? @choochootrain 😄

choochootrain: discussed offline-but-actually-online-because-slack - i think if we defer html escaping to react then we can be p confident that future bugs will have a lot more eyeballs on them :)

erikshestopal:
@kylealwyn ready for final pass.

src/Linkify/Linkify.spec.js
```diff
@@ -10,7 +10,12 @@ describe('Linkify', () => {
   });
 
   test('escapes HTML entities', () => {
-    const component = renderWithTheme(<Linkify>{`<span onmouseover=alert('XSS')></span>`}</Linkify>);
+    const component = renderWithTheme(
+      <Linkify>
+        <img src="fake.jpg" onError={() => {}} alt="hacker" />
```
choochootrain: shouldn't this all be in a template string to test escaping?

src/Linkify/Linkify.js
```diff
@@ -1,19 +1,46 @@
...
-const REGEX = /(\b(https?|ftp|file):\/\/[-A-Z0-9+&@#\/%?=~_|!:,.;]*[-A-Z0-9+&@#\/%=~_|])/gi;
+const detectLinks = ({ children, linkStyle = {} }) =>
+  React.Children.map(children, child => {
+    if (!child) return null;
 
-const parse = string => string.replace(REGEX, url => `<a target="_blank" href="${url}">${url}</a>`);
+    if (typeof child === 'string') {
+      const words = child.split(' ');
+      return words.map((word, index) => {
+        const isLastWord = words.length === index - 1;
```
kylealwyn: copy pasta from my example but this should be `words.length - 1 === index`

erikshestopal:
@kylealwyn @choochootrain ready for final pass.

src/Linkify/Linkify.js
```diff
@@ -1,19 +1,49 @@
 import React from 'react';
-import styled from 'styled-components';
 
-const Container = styled.span`
-  a {
-    color: inherit;
-    text-decoration: underline;
-  }
-`;
+const URL_REGEX = /(\b(https?|ftp|file):\/\/[-A-Z0-9+&@#/%?=~_|!:,.;]*[-A-Z0-9+&@#/%=~_|])/gi;
```
erikshestopal: Do we want to add some rules around domain extensions so that URL's like `duckduckgo.com` in the playground below will also be converted into links?

kylealwyn: In the future, probably yes. We can add rules to our doctors for now :)

erikshestopal: @kylealwyn cool. LGTM 🚢

PR #12, opened 2018-12-18
Allow input to be a number

jlemm45:
Summary:
Allows the form input to accept a number without a warning

### DIFF
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -54,7 +54,7 @@ const StyledTextArea = StyledInput.withComponent('textarea');
 
 export default class Input extends Component {
   static propTypes = {
-    value: PropTypes.string,
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     disabled: PropTypes.bool,
     placeholder: PropTypes.string,
     multiline: PropTypes.bool,


choochootrain:
i would still consider enforcing something stronger here - the more typechecking we can do the better IMO

something like
```js
  ...
  value: function(props, propName, componentName) {
    if (props["type"] === "number" && typeof props[propName] !== "number") {
      return new Error(
        `Invalid prop ${propName} supplied to ${componentName} with type="number", expected Number`
      );
    } else if (typeof props[propName] !== "string") {
      return new Error(
        `Invalid prop ${propName} supplied to ${componentName}, expected String`
      );
    }
  },
  ...
  type: PropTypes.string,
```

jlemm45:
Ah gotcha I see what you mean

src/Form/Input.js
```diff
@@ -52,17 +52,27 @@ const StyledInput = createComponent({
 
 const StyledTextArea = StyledInput.withComponent('textarea');
 
+const validateValueProp = (props, propName, componentName) => {
```
jlemm45: @choochootrain 

Mostly following your example, this should validate our conditions:
If the input has a type of `number` validate the value prop for a `number` type
If the input has a different type validate value prop is a `string` type

src/Form/Input.js
```diff
@@ -52,17 +52,27 @@ const StyledInput = createComponent({
...
+  if (props.type === 'number' && typeof props.value !== 'number') {
```
choochootrain: lets generalize this to `propName` like below: `... && typeof props[propName] !== 'number'`

src/Form/Input.js
```diff
@@ -52,17 +52,27 @@ const StyledInput = createComponent({
...
+    return new Error(`Invalid prop ${propName} supplied to ${componentName} with type="number", expected Number`);
+  }
+  if (typeof props[propName] !== 'string' && props.type !== 'number') {
```
choochootrain: is there a case where you have to add the second check for `props.type !== 'number'`?

jlemm45: Without that this would fail as invalid `<Input value={123} type="number" />`

choochootrain: whoops, good point

PR #15, opened 2019-01-15
fix(Input): Use parentNode.removeChild to teardown textarea shadow node in support of older browsers

kylealwyn:
Also cleans up autogrow logic

### DIFF
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -50,6 +50,15 @@ const StyledInput = createComponent({
   `,
 });
 
+const AutogrowShadowEl = createComponent({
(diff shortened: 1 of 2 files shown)


src/Form/Input.js
```diff
@@ -99,6 +110,15 @@ export default class Input extends Component {
...
+    if (props.value !== undefined && props.value !== state.value) {
+      return {
+        value: props.value,
+      };
+    }
+    return null;
+  }
+
   state = {
     value: this.props.value || '',
```
erikshestopal: Do your changes in `getDerivedStateFromProps` make this unnecessary now?

PR #17, opened 2019-02-14
fix(Dropdown): Support for configuring dropdown offset

cehsu:
The [redesigned EMR dashboard](https://app.asana.com/0/821652619613450/1104820821162262/f) requires the following properties of the dropdown component to be configurable: border-radius, border, offset.

Related CMS PR: https://github.com/sappira-inc/cms/pull/225

Tested locally via `yarn docs`
image

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -15,12 +15,12 @@ const DropdownTrigger = createComponent({
 
 const DropdownMenu = createComponent({
   name: 'DropdownMenu',
-  style: ({ width = 150, theme }) => css`
+  style: ({ width = 150, theme, border, radius }) => css`
     z-index: 10;
     position: absolute;
     background: white;
-    border-radius: ${theme.radius}px;
-    border: 1px solid #e4edf5;
+    border-radius: ${radius || theme.radius}px;
+    border: ${border ? 1px solid #e4edf5 : none};
     box-shadow: 0 0 3px 0 rgba(178, 194, 212, 0.3);
     min-width: 90px;
     width: ${width}px;
@@ -42,6 +42,7 @@ export default class Dropdown extends React.Component {
     placement: 'bottom-start',
     boundariesElement: 'window',
     on: 'click',
+    border: true,
   };
 
   triggerRef = React.createRef();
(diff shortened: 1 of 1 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -71,7 +72,7 @@ export default class Dropdown extends React.Component {
   }
 
   show = () => {
-    const { placement, boundariesElement } = this.props;
+    const { offset = '0, 10', placement, boundariesElement } = this.props;
```
choochootrain: add `offset` to `propTypes` and `defaultProps`?

src/Dropdown/Dropdown.js
```diff
@@ -42,6 +42,7 @@ export default class Dropdown extends React.Component {
     placement: 'bottom-start',
     boundariesElement: 'window',
     on: 'click',
+    border: true,
```
choochootrain: add default for `boxShadow`?

kylealwyn:
Can we just configure the dropdown styles in cms `theme.js`?

cehsu:
Thanks @kylealwyn, that's a better approach -- we can specify the styles in theme instead. An offset will still need to be passed down in order to specify how Popper.js will handle/calculate placement.

src/Dropdown/Dropdown.js
cehsu: Thanks @choochootrain, great idea. Updated.

src/Dropdown/Dropdown.js
```diff
@@ -82,7 +86,7 @@ export default class Dropdown extends React.Component {
           placement,
           modifiers: {
             offset: {
-              offset: '0, 10',
```
kylealwyn: Does this need to be `offset: this.props.offset`?

cehsu: Yes! That was the intention, thanks for catching the oversight, fixed.

kylealwyn: Was this tested locally? Browser would've thrown

cehsu: I hadn't tested it locally, is something like yalc the best tool for that? Or by running the docs command?

cehsu: Thanks for the reminder Kyle, screenshot added to description above. (I had been testing the changes locally by editing the molekule.es.js file in  cms/node_modules, and had forgotten to copy over the assignment.)

kylealwyn: Looks like you found `yarn docs` :)

PR #19, opened 2019-02-22
fix(Dropdown): Close on other dropdown click, configure positionFixed

cehsu:
1. Potentially temporary fix for closing dropdowns when other dropdowns are clicked. 
_We already have support for closing on blur, except when other dropdowns are clicked._

    - Add tabIndex and onBlur to wrapper (to support autoclose on any blur event, including clicking on another dropdown)
    - Add onFocus to menu so that autoclose doesn't get triggered if a blur is caused by focusing on a menu item. This doesn't seem possible currently, as popper.js doesn't support keyboard access, and therefore focus events on its contents. But this behavior is added here in case in the future, we use dropdown menus that allow focus events on their items.

2. Configure `positionFixed`, to support offset

image

[Asana task](https://app.asana.com/0/821652619613450/1110934258569822/f)

[Reference for new approach](https://reactjs.org/docs/accessibility.html#mouse-and-pointer-events)

~As illustrated in the reference, `click` is also made the event on which the dropdown is toggled (rather than being configurable with `click` as a default option), and attached to the trigger directly. There don't appear to be any instances of dropdowns being opened otherwise (via other events). This change can be removed if desired.~

This change is potentially temporary as our dropdowns don't appear to be keyboard accessible currently? ([Keyboard accessibility requirements for dropdowns](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html#kbd_label).) This isn't supported by Popper.js either. An alternative such as Downshift (already being used for `<Autocomplete/>` in the EMR) could be appropriate to use here instead in the future.

There are some changes made to the markdown, which are for reviewer local testing convenience, and which are not intended to be merged.


src/Dropdown/Dropdown.js
```diff
@@ -121,45 +120,42 @@ export default class Dropdown extends React.Component {
...
   render() {
     const { width, trigger, render, children } = this.props;
     const { isOpen } = this.state;
 
     const renderFn = render || children;
 
     return (
-      <Fragment>
-        <DropdownTrigger ref={this.triggerRef} aria-haspopup="true" aria-expanded={isOpen}>
+      <DropdownWrapper tabIndex={0} onBlur={this.handleBlur}>
```
cehsu: The reference does not add a tabIndex here, but I'm not sure how to register focus or blur events on a div otherwise.

src/Dropdown/Dropdown.js
```diff
@@ -121,45 +120,42 @@ export default class Dropdown extends React.Component {
...
+        <DropdownTrigger onClick={this.toggle} ref={this.triggerRef} aria-haspopup="true" aria-expanded={isOpen}>
           {trigger}
         </DropdownTrigger>
 
         <Portal>
           {isOpen && (
-            <DropdownMenu ref={this.menuRef} width={width}>
+            <DropdownMenu ref={this.menuRef} width={width} onFocus={this.handleFocus} tabIndex={0}>
```
cehsu: Since our dropdown menu is managed by popper.js, attaching the onFocus handler here helps to register focus events properly (it was not firing when attached to the DropdownWrapper).

src/Dropdown/Dropdown.mdx
```diff
@@ -16,6 +16,28 @@ Easily display contextual overlays using custom trigger elements. Dropdown's pos
...
+
+          <Dropdown.SectionTitle>Section Two</Dropdown.SectionTitle>
+          <Dropdown.Item onClick={close}>Item One</Dropdown.Item>
+          <Dropdown.Item disabled>Item Two</Dropdown.Item>
+        </Dropdown.Body>
+
+        <Dropdown.Footer>Footer</Dropdown.Footer>
+      </>
+    )}
+  </Dropdown>
```
cehsu: We probably don't want to add another example, but this is available to facilitate local testing.

src/Dropdown/Dropdown.js
```diff
@@ -121,45 +120,42 @@ export default class Dropdown extends React.Component {
...
     const { width, trigger, render, children } = this.props;
     const { isOpen } = this.state;
 
     const renderFn = render || children;
 
     return (
-      <Fragment>
-        <DropdownTrigger ref={this.triggerRef} aria-haspopup="true" aria-expanded={isOpen}>
+      <DropdownWrapper tabIndex={0} onBlur={this.handleBlur}>
+        <DropdownTrigger onClick={this.toggle} ref={this.triggerRef} aria-haspopup="true" aria-expanded={isOpen}>
```
cehsu: Do we want to clone the `trigger` and pass it these props instead ([example](https://stackoverflow.com/a/32371612)) so we can avoid using this wrapper div?

kylealwyn: Sure - if it continues to work as intended. There may be an issue with passing the ref to `trigger`

src/Dropdown/Dropdown.js
```diff
@@ -121,45 +122,42 @@ export default class Dropdown extends React.Component {
...
-
   autoclose = e => {
-    if (this.state.isOpen && !this.menuRef.current.contains(e.target)) {
+    if (this.state.isOpen) {
       this.hide();
     }
   };
 
+  handleBlur = () => {
+    this.timeoutId = setTimeout(this.autoclose, 500);
```
kylealwyn: 500 feels a little long in testing - let's go for 175 or so.

src/Dropdown/Dropdown.js
```diff
@@ -1,11 +1,18 @@
...
 import { css } from 'styled-components';
 import Popper from 'popper.js';
 import Box from '../Box';
 import Portal from '../Portal';
 import { createComponent, themeGet } from '../utils';
 
+const DropdownWrapper = createComponent({
+  name: 'DropdownWrapper',
+  style: css`
+    display: inline-block;
```
kylealwyn: This could be considered a breaking change so let's think through if this could have unintended consequences anywhere we currently use Dropdown

src/Dropdown/Dropdown.js
cehsu: Actually since trigger is a JSX template literal (and not an element) this is best left as is.

src/Dropdown/Dropdown.js
cehsu: Delay removed.

src/Dropdown/Dropdown.js
cehsu: Due to the event configuration being removed (on line 47, etc.)?

cehsu: If so, I could not find any instances of the `on` prop being used to specify an alternative event listener.

kylealwyn: Due to adding the new `inline-block` style to the wrapper, Dropdown now becomes inline with their previous sibling, which may have unintended consequences anywhere we use the component.

## Before change
screen shot 2019-03-01 at 9 43 24 am

## After change
screen shot 2019-03-01 at 9 43 29 am

---

Regarding `on` change, we have a ticket coming up that will require showing the dropdown on hover. I'd prefer not to move or if we need to make changes to support accessibly in this ticket, let's do that in this PR: https://app.asana.com/0/1107902406623186/1112247428608998

src/Dropdown/Dropdown.js
```diff
@@ -121,45 +122,42 @@ export default class Dropdown extends React.Component {
...
+    if (this.state.isOpen) {
       this.hide();
     }
   };
 
+  handleBlur = () => {
+    this.timeoutId = setTimeout(this.autoclose);
+  }
+
+  handleFocus = () => {
```
kylealwyn: Is this necessary anymore now that we removed the blur timeout?

src/Dropdown/Dropdown.js
cehsu: Many thanks @kylealwyn, will remove the styling change, and revert the removal of the event configuration.

src/Dropdown/Dropdown.js
cehsu: Since the dropdowns cannot receive focus, we can remove this.

src/Dropdown/Dropdown.js
```diff
@@ -143,7 +136,7 @@ export default class Dropdown extends React.Component {
     const renderFn = render || children;
 
     return (
-      <DropdownWrapper tabIndex={0} onBlur={this.handleBlur}>
+      <div tabIndex={0} onBlur={this.handleBlur}>
```
choochootrain: did you mean to rm `DropdownWrapper`? looks like you removed `DropdownTrigger` which is still in use

cehsu: Thanks @choochootrain, fixed.

src/Dropdown/Dropdown.js
cehsu: It turns out we need this, after testing locally. The `setTimeout` doesn't need an interval since it's purpose is to perform the callback on the next tick (after `handleFocus` has fired, if necessary). I've added comments about this.

src/Dropdown/Dropdown.js
cehsu: 175 delay restored for mouseenter/mouseleave support.

src/Dropdown/Dropdown.js
```diff
@@ -111,55 +116,67 @@ export default class Dropdown extends React.Component {
...
     } else {
       this.show();
     }
   };
 
-  attachAutocloseListener() {
-    if (!this.props.autoclose) {
-      return;
-    }
-    document.addEventListener('click', this.autoclose);
```
cehsu: It seems like we can rely on React's supported synthetic events here and not need to use `addEventListener`, which is generally discouraged in the [docs](https://reactjs.org/docs/handling-events.html).

> When using React you should generally not need to call addEventListener to add listeners to a DOM element after it is created. Instead, just provide a listener when the element is initially rendered.

src/Dropdown/Dropdown.js
```diff
@@ -111,55 +116,67 @@ export default class Dropdown extends React.Component {
...
-  }
-
-  detachAutocloseListener() {
-    document.removeEventListener('click', this.autoclose);
-  }
-
   autoclose = e => {
-    if (this.state.isOpen && !this.menuRef.current.contains(e.target)) {
+    if (this.state.isOpen && this.props.autoclose) {
```
cehsu: It doesn't seem like this prop was being used before.

src/Dropdown/Dropdown.js
```diff
@@ -111,55 +116,67 @@ export default class Dropdown extends React.Component {
...
   render() {
     const { width, trigger, render, children } = this.props;
     const { isOpen } = this.state;
 
     const renderFn = render || children;
 
     return (
-      <Fragment>
-        <DropdownTrigger ref={this.triggerRef} aria-haspopup="true" aria-expanded={isOpen}>
+      <div tabIndex={0} onBlur={this.handleBlur}>
```
cehsu: This needs to be a `div` to specify a `tabIndex`, `onBlur`.

src/Dropdown/Dropdown.js
cehsu: In order to support accessibility, we would likely need to switch from popper.js to something like downshift, as discussed [here](https://app.asana.com/0/821652619613450/1110934258569822/f).

src/Dropdown/__snapshots__/Dropdown.spec.js.snap
```diff
@@ -42,16 +42,24 @@ exports[`Dropdown 1`] = `
...
-    size="md"
+  <div
+    aria-expanded={false}
+    aria-haspopup="true"
+    className="re-dropdown-trigger c0"
+    onClick={[Function]}
+    onMouseEnter={[Function]}
   >
-    Trigger
-  </button>
```
cehsu: It seems like more tests would be a good idea here? I will add some more.

choochootrain: can we add snapshots after the dropdown has been opened/closed to test event propagation?

cehsu: This seems possible: https://jestjs.io/docs/en/tutorial-react#snapshot-testing

src/Dropdown/Dropdown.js
cehsu: @kylealwyn do we still want to support hover if the upcoming ticket has changed to be click-based?

kylealwyn:
screen shot 2019-03-04 at 3 11 02 pm Could we remove the outline on the trigger div?

src/Dropdown/__snapshots__/Dropdown.spec.js.snap
cehsu: It turns out we can use our existing dependencies for this. [Another branch](url) has been set up for that work (r[elevant commit](https://github.com/sappira-inc/molekule/commit/80697576237ec957efc0d27f302bf2b1fd927316)), so that it doesn't block this.

src/Dropdown/Dropdown.js
```diff
@@ -1,15 +1,28 @@
...
+  click: 'focus',
+  mouseenter: 'mouseenter',
+};
+
 const DropdownTrigger = createComponent({
   name: 'DropdownTrigger',
   style: css`
     display: inline-block;
+    &:focus {
+      outline: none;
```
cehsu: This is an accessibility no no: 
> Buttons need a fully visible keyboard focus outline so users can see them when tabbing.
https://www.deque.com/blog/accessible-aria-buttons/

Should we create an accessibility audit task, and include issues like this as sub-tasks?

choochootrain: i think thats a good diea - there's probably a lot of accessibility oopsies to be addressed anyways

src/Dropdown/__snapshots__/Dropdown.spec.js.snap
choochootrain: do you want to roll it into this PR or is this otherwise ready to go?

cehsu: I've set up a [PR](https://github.com/sappira-inc/molekule/pull/25) to roll it into this PR if that's alright?

kylealwyn:
Lived on in https://github.com/heydoctor/molekule/pull/27

PR #20, opened 2019-02-22
feat(Formbot): support for yup via validationSchema; Add Context provider and EasyInput consumer

choochootrain:
* supports `yup` schemas for validation:
  pulled from EMR `Formbot` - `validationSchema` takes precedence over `validations`
* manages state of `Inputs` within `Formbot` context:
  see #21 for details
* minor addition of `Fieldset` component:
  pulled from EMR


kylealwyn:
Let's add docz example using yup

src/Form/Formbot.js
```diff
@@ -35,6 +38,7 @@ export default class Formbot extends React.Component {
   static defaultProps = {
     initialValues: {},
     validations: {},
+    validationSchema: {},
```
cehsu: Should this have a `fieldName` property?

src/Form/Formbot.js
```diff
@@ -48,7 +52,7 @@ export default class Formbot extends React.Component {
   };
 
   get validatable() {
-    return Object.keys(this.props.validations);
+    return Object.keys(this.props.validationSchema || this.props.validations).length !== 0;
```
cehsu: If we want to get terse, we can rewrite this as:
`!!Object.keys(this.props.validationSchema || this.props.validations).length`

src/Form/Formbot.js
```diff
@@ -110,15 +114,27 @@ export default class Formbot extends React.Component {
 
   validateField(field) {
     return new Promise(resolve => {
-      const validation = this.props.validations[field];
+      const fieldState = this.state.fields[field] || {};
```
cehsu: Could we use lodash `get` to simplify this?

src/Form/Formbot.js
```diff
@@ -110,15 +114,27 @@ export default class Formbot extends React.Component {
...
+      if (fieldState.validated) {
+        resolve();
+        return;
+      }
 
-      if (!validation) return;
+      const fromSchema = !!this.props.validationSchema;
+      const validation = (fromSchema ? this.props.validationSchema : this.props.validations)[field];
```
cehsu: Could these two lines be condensed into `(this.props.validationSchema || this.props.validations)[field]`?

src/Form/Formbot.js
```diff
@@ -110,15 +114,27 @@ export default class Formbot extends React.Component {
...
+        resolve();
+        return;
+      }
 
       const fieldValue = this.state.values[field];
       let errorMsg;
 
       try {
-        if (typeof validation === 'function') {
+        if (fromSchema) {
```
cehsu: I see we're reusing `fromSchema` here so that we don't have to perform the `!!this.props.validationSchema` check twice. If this is the only instance, can we perform the check inline?

src/Form/Formbot.js
```diff
@@ -132,7 +148,11 @@ export default class Formbot extends React.Component {
           });
         }
       } catch (err) {
-        errorMsg = err.message;
+        if (fromSchema) {
```
cehsu: Here is is again, let's keep it in that case.

src/Form/Formbot.js
```diff
@@ -132,7 +148,11 @@ export default class Formbot extends React.Component {
...
+          errorMsg = err ? err.errors[0] : undefined;
```
cehsu: Is it possible for `err` to be falsy, if we're inside catch?

src/Form/FormGroup.js
```diff
@@ -1,6 +1,36 @@
...
+  `,
+});
+
+const Container = styled(Box)`
+  * + & {
+    margin-top: 1.5rem;
+  }
+`;
+
+const FormGroup = ({ legend, children, padding = 3 }) => (
```
kylealwyn: To keep with underlying html semantics, let's call this fieldset

kylealwyn:
screen shot 2019-02-22 at 5 05 27 pm
After clicking submit in new yup example

src/Form/Formbot.js
cehsu: Is this only used to map? Do we just want to return the keys?

src/Form/Formbot.js
choochootrain: just porting it over from the EMR but yeah probably not

src/Form/Formbot.js
choochootrain: it looks like we're not using lodash at all so i dont want to add it just for this instance

src/Form/FormGroup.js
choochootrain: thats gonna require changing all the FormGroup instances in the emr and web app - we can export it under both names for now?

src/Form/Formbot.js
choochootrain: no, the proptypes aren't doing the right thing so i'll just rm

src/Form/Formbot.example.js
```diff
@@ -68,7 +68,7 @@ export default function() {
       }}>
       {({ values, onSubmit, onChange, errors, onBlur }) => (
         <form onSubmit={onSubmit}>
-          <FormGroup>
+          <FormGroup legend="A Group of Inputs">
```
cehsu: If `values.name` is `undefined`, I think the input still expects the `value` to be of type `string`, so we will likely want to set the value to empty string if it is undefined? Othewise a proptype error results.

src/index.js
```diff
@@ -16,6 +16,7 @@ export Flex from './Flex';
 export Formbot from './Form/Formbot';
 export FormError from './Form/FormError';
 export FormGroup from './Form/FormGroup';
+export FieldSet from './Form/FormGroup';
```
kylealwyn: `Fieldset` should be an entirely new component. FormGroups refer to https://getbootstrap.com/docs/4.0/components/forms/#form-groups which adds bottom margin to adjacent inputs

src/Form/Formbot.example.js
cehsu: It looks like the message field doesn't specify the value prop, is that an option here?

src/index.js
choochootrain: added but `FieldSet` does something similar in the presence of adjacent elements - unclear how one would mix usage of  `FieldSet` and `FormGroup`

choochootrain: hm also `web-app` has a different `FieldSet` which has a sublegend and is styled differently - should we leave them as is or theme `FieldSet`s?

choochootrain:
theres also `PhoneInput`, `DateInput` in web-app so this PR will get pretty big

i can ship this as is and update the EMR and we can do a second pass to clean up the web-app next hack day

kylealwyn:
@choochootrain 

Supernit: Can we rename FieldSet to Fieldset? HTML tag reads `fieldset` so I'd prefer sticking to convention and start casing?

---

> added but `FieldSet` does something similar in the presence of adjacent elements - unclear how one would mix usage of `FieldSet` and `FormGroup`

Our FormGroup and Field are very similar. Most (should be all) of our input components are wrapped in `<Field />`, which adds top margin to siblings. FormGroup, on the other hand, always applies a bottom margin.

Fieldset is used for optionally grouping related fields if multiple sets exist and conventionally has the `legend` element as first descendant. See `edit-profile.js` in web app, where we group personal and address information. 

The css class `form-group` in Bootstrap always applies a margin bottom, and we use it to wrap components in a form which may not have a margin applied, or most commonly, a row of inputs within columns where the css sibling selector breaks. This is more a sugar component for consistency, as we could do `<Box mb={3}>` and get the same results

```js
<Form>
  <Fieldset legend="Some group of fields">
    <Input />
    <Input /> {/* Here we have adjacent Fields so this one has a top margin applied */}
  </Fieldset>

  <Fieldset legend="Some other group of fields">
    <FormGroup>
      <Row>
         <Col>
            <Input />
         </Col>
         <Col>
            <Input />
         </Col>
      </Row>
    </FormGroup>
  </Fieldset>

  <Button type="submit" />
</Form>
```

---

> theres also `PhoneInput`, `DateInput` in web-app so this PR will get pretty big
> 
> i can ship this as is and update the EMR and we can do a second pass to clean up the web-app next hack day

This is the right call.

choochootrain:
@kylealwyn looks like `FormGroup` uses `Box` in favor of bootstrap's `form-group` class

```js
const FormGroup = ({ children, padding = 3 }) => <Box mb={padding}>{children}</Box>;
```

kylealwyn:
@choochootrain yeah, was only using bootstrap as reference for why it exists

src/index.js
```diff
@@ -12,7 +12,7 @@ export Column from './Grid/Column';
 export Container from './Grid/Container';
 export Dropdown from './Dropdown';
 export Field from './Form/Field';
-export FieldSet from './Form/FieldSet';
+export Fieldset from './Form/Fieldset';
```
kylealwyn: Let's rename the file too - would cause issues buliding on OSs like Linux that care about path capitalization

choochootrain:
@sappira-inc/engineering take a look at the last change, i'm open to suggestions

PR #21, opened 2019-02-23
Add Context to Formbot with EasyInput consumer and HOC

kylealwyn:


src/Form/EasyInput.js
```diff
@@ -0,0 +1,34 @@
...
+import { Context } from './Formbot';
+
+/**
+ * useContext has some slight performance issues. Every update to context re-renders
+ * each subscribed component and there's currently no way to bail out of renders if the
+ * values we care about haven't changed. Below is a sufficient workaround below and redux maintainers are discussing here: https://github.com/facebook/react/issues/14110
+ */
+const PureInput = React.memo(({ Component, ...props }) => <Component {...props} />);
+
+function EasyInput({ name, Component, ...props  }) {
```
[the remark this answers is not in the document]
kylealwyn: We're not preserving scope, not implicitly returning and an arrow wouldn't reduce verbosity, so I'd argue that function keyword is actually more semantically correct here.

However, will move to arrow for consistency as you mentioned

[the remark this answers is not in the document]
kylealwyn: screen shot 2019-02-25 at 11 27 36 am

Personally, I think `function Component` reads better. I'd prefer that we follow the rules I mentioned above, which may create the consistency you're looking for. Let's discuss further if you want

src/Form/Formbot.example.js
```diff
@@ -72,80 +80,40 @@ export default function() {
...
-          </Fieldset>
-          <FormGroup>
-            <Switch name="switch1" id="switch1" onChange={onChange} toggled={values.switch1} />
-          </FormGroup>
-          <Button htmlType="submit" type="primary" size="sm">
-            Submit
-          </Button>
-        </form>
-      )}
+      <Form>
```
choochootrain: 🚮

src/Form/Input.js
```diff
@@ -277,3 +278,5 @@ export default class Input extends Component {
     );
   }
 }
+
+export default createEasyInput(Input);
```
choochootrain: do you intend for `Input`s to be overridable? e.g.
```js
<Input name="my_field" value={some_function(values.my_field)} />
```

looks like the HOC will take precedence

kylealwyn: Yeah - props are spread last so any auto-bound prop can be overridden. Seems like the right behavior but do you have a case in mind where we wouldn't want that? 
```diff
  if (!state) {
    return <PureInput name={name} Component={Component} {...props} />
  }

   return (
    <PureInput
      name={name}
      value={state.values[name] || ''}
      error={state.errors[name]}
      onChange={state.onChange}
      onBlur={state.onBlur}
      onFocus={state.onFocus}
      Component={Component}
+    {...props}
    />
  );
```

choochootrain: no im just dumb, i missed the spread

PR #22, opened 2019-02-26
fix: Forward input refs, add missing form export, remove autogrow shadow element from tab order

choochootrain:


erikshestopal:
Would it be too much to ask for you to add a `tabIndex={-1}` to the shadow input [here](https://github.com/sappira-inc/molekule/blob/master/src/Form/Input.js#L56)?
This shadow input messes up tabbing to the next input if the first `Input` component has `autogrow` as a prop. 

If you plop this code into the molekule docs playground, you'll see that it takes two tab presses to get to the next input element. 
```jsx
<Input
  multiline
  autogrow
  placeholder="Example"
  label="Example"
  value="I already have a value! If provided a value, I become controlled, so I won't update when you type here..."
/>

<Input placeholder="Example" label="Example"  />
```

PR #24, opened 2019-03-02
fix(Modal): Use Transition to simplify animation; add close button; convert to hooks

kylealwyn:


src/Modal/Modal.js
```diff
@@ -38,163 +49,116 @@ const Backdrop = createComponent({
...
-    this.closeInternal();
-  };
-
-  handleContentClick = event => {
-    event.stopPropagation();
-  };
+  useEffect(() => {
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  })
```
choochootrain: worth conditionally firing with `[isOpen, props.closeOnEscape]`?

kylealwyn: Just gave that a try. Presented weird behavior that closes both modals and reopens the second when pressing escape

PR #25, opened 2019-03-08
Add functional Dropdown tests

cehsu:
## Description
These changes add functional tests for the Dropdown component.

In order to support the use of async/await, `.babelrc` had to be configured according to the [Jest docs](https://jestjs.io/docs/en/getting-started#using-babel).

## Task
https://app.asana.com/0/887402852146010/1112746503758429/f

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -16,14 +16,11 @@
   "env": {
     "test": {
       "presets": [
-        [
-          "@babel/env",
-          {
-            targets: {
-              node: 'current',
-            },
-          },
-        ]
+        ["@babel/env", {
+          "targets": {
+            "node": "current"
+          }
+        }]
       ]
     }
   }
--- a/.babelrc
+++ b/.babelrc
@@ -15,13 +15,7 @@
   ],
   "env": {
     "test": {
-      "presets": [
-        ["@babel/env", {
-          "targets": {
-            "node": "current"
-          }
-        }]
-      ]
+      "presets": ["@babel/env"]
     }
   }
 }
--- a/.babelrc
+++ b/.babelrc
@@ -15,7 +15,16 @@
   ],
   "env": {
     "test": {
-      "presets": ["@babel/env"]
+      "presets": [
+        [
+          "@babel/env",
+          {
+            targets: {
+              node: 'current',
+            },
+          },
+        ]
+      ]
     }
   }
 }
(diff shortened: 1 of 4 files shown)


.babelrc
```diff
@@ -15,7 +15,13 @@
   ],
   "env": {
     "test": {
-      "presets": ["@babel/env"]
+      "presets": [
+        ["@babel/env", {
+          "targets": {
+            "node": "current"
```
choochootrain: why do we need this?

src/Dropdown/Dropdown.spec.js
```diff
@@ -24,3 +42,79 @@ test('Dropdown', () => {
...
+      )}
+    </Dropdown>);
+  
+  // open on trigger
+  const trigger = wrapper.childAt(0);
+  trigger.simulate(event, { type: event, stopPropagation });
+
+  // blur dropdown
+  wrapper.simulate(blurEvent, { type: blurEvent, stopPropagation });
+  await setTimeout(() => {
```
choochootrain: nit: not sure `await setTimeout` does what you want and you're using the `done` callback anyways

src/Dropdown/Dropdown.spec.js
```diff
@@ -24,3 +42,79 @@ test('Dropdown', () => {
...
+
+  // open on trigger
+  const trigger = wrapper.childAt(0);
+  trigger.simulate(event, { type: event, stopPropagation });
+
+  // blur dropdown and focus on menu
+  const menu = wrapper.childAt(1).childAt(0);
+  wrapper.simulate(blurEvent, { type: blurEvent, stopPropagation });
+  menu.simulate(focusEvent, { type: focusEvent, stopPropagation });
+  await setTimeout(() => {
```
choochootrain: same as above

src/Dropdown/Dropdown.spec.js
cehsu: Good point, removed.

src/Dropdown/Dropdown.spec.js
cehsu: Thanks, removed.

.babelrc
cehsu: This was added per the [Jest setup docs Babel guidance](https://jestjs.io/docs/en/getting-started#using-babel), specifically to support the use of async/await. Since we're no longer using async/await, this change has been reverted.

kylealwyn:
Lived on in https://github.com/heydoctor/molekule/pull/27

PR #26, opened 2019-03-08
fix(Card): Wrap in React.forwardRef

cehsu:
## Description
This change wraps `<Card/>` in `React.forwardRef`.

## Task
https://app.asana.com/0/887402852146010/1112252089781752/f

### DIFF
--- a/src/Card/Card.js
+++ b/src/Card/Card.js
@@ -16,7 +16,7 @@ const StyledCard = createComponent({
   `,
 });
 
-const Card = props => <StyledCard {...props} />;
+const Card = React.forwardRef((props, ref) => <StyledCard ref={ref} {...props} />);
 
 Card.propTypes = {
   shadow: PropTypes.bool,


src/Card/Card.js
```diff
@@ -16,7 +16,7 @@ const StyledCard = createComponent({
   `,
 });
 
-const Card = props => <StyledCard {...props} />;
+const Card = React.forwardRef((props, ref) => <StyledCard ref={ref} {...props} />);
```
choochootrain: `styled-components` will forward refs for you but if you want the ref to be on `Box` instead of the wrapped `StyledCard` you might have to verify that `createComponent` does the right thing. could be worth adding a test in that case

cehsu: Thanks, I will add a test to ensure that the ref is assigned as expected.

kylealwyn: This is right - functional components don't forward refs out of the box and this ref should be passed directly to Box. We're wrapping some components as a [hack](https://github.com/pedronauck/docz/issues/156#issuecomment-411002811) for the <PropsTable /> component to work for styled-components. All documentation generators (storybook, docz, stylguidist) are affected due to the issue in the underlying react-docgen parser. Searching for better solutions

PR #27, opened 2019-03-13
feat(Dropdown): Make Dropdown accessible

kylealwyn:
@cehsu gave me the accessibility bug. This adds a focus trap to Dropdowns which forces tabbing to stay trapped within the dropdown menu when open.

Kapture 2019-03-13 at 10 39 59


src/Dropdown/Dropdown.js
```diff
@@ -1,189 +1,216 @@
...
+                boundariesElement,
+                padding: 12,
+              },
+              // Prevents wacky jumps in Transition component
+              computeStyle: { gpuAcceleration: false },
+            }}>
+            {({ ref, style }) => (
+              <Transition in={isOpen} timeout={0} appear>
+                {state => (
+                  <FocusTrap autoFocus={false}>
```
cehsu: Woot! This is a big accessibility win!

In the future, as mentioned in [the dropdown ticket](https://app.asana.com/0/821652619613450/1110934258569822/f), the defined behavior for dropdowns is to be cycled through via arrow keys, toggle via enter, and tab should continue to cycle through all other tabbale elements on the page (I had to build a vanilla html css custom dropdown in the past, which is why this is seared in my memory), as demonstrated here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html, and as enumerated here: https://www.w3.org/TR/wai-aria-practices-1.1/#listbox_kbd_interaction.

Downshift provides support for the specified behaviors, should we consider using that in the future to handle all of this?

kylealwyn: That's a great example but I'd argue our Dropdown is providing different functionality than what looks to be more of custom "select" box. The user engaging in an active action rather than choosing a value. The listbox above is most likely part of a form, which I believe is why tab would move to the next element or input. Conceptually, I think the dropdown is most like a local modal, so I think we'd want to abide by the criteria laid out in https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal 

As precedent, I think Twitter is one of the most accessible big box sites and I know they've poured a grand amount of time, attention and money into their accessbility. Below is an example of tabbing and arrowing through a dropdown.
Kapture 2019-03-13 at 12 07 03

cehsu: Thanks Kyle, you're right! A nav menu is a different set of criteria from a listbox.
Here are the [accessibility specifications](https://www.w3.org/TR/wai-aria-practices/#menu):
[An example](https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html#)
[Another example](https://www.w3.org/WAI/tutorials/menus/application-menus-code/)

kylealwyn: It seems that those examples don't allow tabbing to any element in the menu. I prefer to follow Google/Twitter/Slack precedent which allows tabbing to dropdown/popover items and keep tabbing/arrowing locked within the box.

PR #28, opened 2019-03-15
(fix)RadioGroup: Size prop mismatch, configure style

cehsu:
These changes fix a bug where `<RadioGroup/>` accepted a `size` (string) prop, but it's child component `<Checkbox/>` only accepted `iconSize` (number), and `fontSize` (number). Also styles are now passed down to support more customization (margins, etc.), such as required by [EMR Message Conversation List](https://app.asana.com/0/821652619613450/1108680987294666/f).

image

### DIFF
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -61,7 +61,7 @@ class Checkbox extends React.Component {
     color: PropTypes.string,
     horizontal: PropTypes.bool,
     disabled: PropTypes.bool,
-    styles: PropTypes.object,
+    styles: PropTypes.shape(),
   };
 
   static defaultProps = {
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -145,7 +145,7 @@ class Checkbox extends React.Component {
         <Flex alignItems="center">
           <StyledIcon size={iconSize} color={checked ? colorOn : colorOff} name={checked ? iconOn : iconOff} />
 
-          {label && <StyledLabel fontSize={fontSize}>{label}</StyledLabel>}
+          {label && <StyledLabel fontSize={fontSize} style={styles.Label}>{label}</StyledLabel>}
         </Flex>
 
         {!this.state.focused && error ? <FormError>{error}</FormError> : null}
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -19,6 +19,7 @@ const CheckboxContainer = createComponent({
       margin-left: ${p => (p.horizontal ? '12px' : 0)};
       margin-top: ${p => (p.horizontal ? 0 : '4px')};
     }
+    ${p => p.style}
   `,
 });
 
@@ -28,14 +29,6 @@ const StyledInput = createComponent({
   style: css`
     display: none;
     pointer-events: ${p => (p.disabled ? 'none' : 'auto')};
-  `,
-});
-
-const StyledIcon = createComponent({
-  name: 'CheckboxIcon',
-  as: Icon,
-  style: ({ margin }) => css`
-    margin: ${margin};
   `,
 });
 
@@ -64,6 +57,7 @@ class Checkbox extends React.Component {
     color: PropTypes.string,
     horizontal: PropTypes.bool,
     disabled: PropTypes.bool,
+    styles: PropTypes.object,
   };
 
   static defaultProps = {
@@ -78,6 +72,7 @@ class Checkbox extends React.Component {
     horizontal: false,
     onChange() {},
     disabled: false,
+    styles: {},
   };
 
   static getDerivedStateFromProps(props, state) {
@@ -123,16 +118,16 @@ class Checkbox extends React.Component {
       iconOn,
       iconOff,
       iconSize,
-      iconMargin,
       colorOn,
       colorOff,
       horizontal,
       disabled,
+      styles,
     } = this.props;
     const { checked } = this;
 
     return (
-      <CheckboxContainer horizontal={horizontal}>
+      <CheckboxContainer horizontal={horizontal} style={styles.CheckboxContainer}>
         <StyledInput
           id={id}
           name={name}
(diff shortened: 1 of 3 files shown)


src/Form/Checkbox.js
```diff
@@ -19,6 +19,7 @@ const CheckboxContainer = createComponent({
       margin-left: ${p => (p.horizontal ? '12px' : 0)};
       margin-top: ${p => (p.horizontal ? 0 : '4px')};
     }
+    ${p => p.style}
```
cehsu: Given how the margins are set above, this seemed like the least ornate (e.g. specifying a `margin` prop, which I considered earlier), but still effective option. Thanks for any thoughts on whether this isn't intuitive or consistent.

choochootrain: for those (like me) who dont understand this dark magic:
https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/constructors/css.js
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates

kylealwyn: I don't think line is necessary. styled-components take a style prop by default

src/Form/RadioGroup.js
```diff
@@ -60,7 +63,7 @@ class RadioGroup extends Component {
   };
 
   render() {
-    const { choices, error, horizontal, label, name, colorOn, colorOff, size } = this.props;
+    const { choices, error, horizontal, label, name, colorOn, colorOff, fontSize, iconSize, styles } = this.props;
```
kylealwyn: we could spread these final five props into something like `...checkboxProps` and spread onto `<Checkbox />` below

src/Form/RadioGroup.js
```diff
@@ -87,7 +90,9 @@ class RadioGroup extends Component {
                   valueFalse={value}
                   iconOn="radiobox-marked"
```
kylealwyn: Can we add `iconOn` and `iconOff` as props while we're here and set defaultProps to these values?

src/Form/Checkbox.js
cehsu: Hm, I tried removing this line and the style no longer got applied.

src/Form/RadioGroup.js
cehsu: Sounds good, will update.

src/Form/RadioGroup.js
cehsu: Sounds good, will update.

src/Form/RadioGroup.js
cehsu: Thanks, updated.

src/Form/RadioGroup.js
cehsu: Thanks, updated.

src/Form/Checkbox.js
kylealwyn: It looks like it's an issue with the test style in the doc. If you exchange `"margin: 0 !important"` for 0, it works without this line

src/Form/RadioGroup.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { css } from 'styled-components';
```
kylealwyn: unused

src/Form/RadioGroup.js
```diff
@@ -20,20 +21,26 @@ class RadioGroup extends Component {
...
     choices: PropTypes.arrayOf(
       PropTypes.shape({
         value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
         label: PropTypes.string,
         disabled: PropTypes.bool,
       })
     ),
+    styles: PropTypes.object,
+    iconOn: PropTypes.string,
+    iconOf: PropTypes.string,
```
kylealwyn: typo

src/Form/RadioGroup.js
```diff
@@ -20,20 +21,26 @@ class RadioGroup extends Component {
...
+    fontSize: PropTypes.number,
+    iconSize: PropTypes.number,
     choices: PropTypes.arrayOf(
       PropTypes.shape({
         value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
         label: PropTypes.string,
         disabled: PropTypes.bool,
       })
     ),
+    styles: PropTypes.object,
```
kylealwyn: prefer shape() to make linter happy

src/Form/RadioGroup.js
```diff
@@ -60,7 +67,7 @@ class RadioGroup extends Component {
   };
 
   render() {
-    const { choices, error, horizontal, label, name, colorOn, colorOff, size } = this.props;
+    const { choices, error, horizontal, label, name, colorOn, colorOff, ...checkboxProps } = this.props;
```
kylealwyn: colorOn and colorOff and be swept into checkboxProps spread

src/Form/Checkbox.js
```diff
@@ -141,7 +145,7 @@ class Checkbox extends React.Component {
         <Flex alignItems="center">
           <StyledIcon size={iconSize} color={checked ? colorOn : colorOff} name={checked ? iconOn : iconOff} />
 
-          {label && <StyledLabel fontSize={fontSize}>{label}</StyledLabel>}
+          {label && <StyledLabel fontSize={fontSize} style={styles.Label}>{label}</StyledLabel>}
```
cehsu: This style is applied without specifying it in the styled component's template literal, I'm not sure why the other style does not.

kylealwyn: https://github.com/heydoctor/molekule/pull/28/files#r266171170

src/Form/Checkbox.js
cehsu: Oh thanks, I'll do that.

src/Form/RadioGroup.js
cehsu: Thanks, replaced with `shape()`.

src/Form/RadioGroup.js
cehsu: OK, removed.

src/Form/RadioGroup.js
cehsu: Thanks, fixed.

src/Form/RadioGroup.js
cehsu: Thanks, removed.

PR #33, opened 2019-03-22
(fix)Dropdown: Configure <Dropdown.Trigger/> style

cehsu:
These changes allow the `<Dropdown/>` component's `<TriggerWrapper/>`(`styled(Dropdown.Trigger)`) to be styled so that it may, for example, fill the available width of its parent container.
image

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -28,7 +28,7 @@ export const PLACEMENT_TRANSITION_ORIGINS = {
 
 const ARROW_KEYS = ['ArrowUp', 'ArrowDown'];
 
-const TriggerWrapper = createComponent({
+const Trigger = createComponent({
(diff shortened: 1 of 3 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -114,7 +115,7 @@ export default function Dropdown({
       <Manager>
         <Reference>
           {({ ref: triggerRef }) => (
-            <TriggerWrapper ref={triggerRef} tabIndex={-1}>
+            <TriggerWrapper style={styles.TriggerWrapper} ref={triggerRef} tabIndex={-1}>
```
kylealwyn: Nice - can we call this `styles.Trigger` and change component to `<Trigger />` to reduce verbosity

cehsu: Good idea.

PR #35, opened 2019-03-26
fix(Icon): Prevent passed className from being overwritten

erikshestopal:
## Summary 
When the `Icon` component was being provided with a `className` prop, this prop was overwriting the default class name generated for the icon, namely `mdi mdi-X`. In the web-app several icons disappeared mysteriously and this was because those component had this structure...
```js
<Icon className="blah" name="message" />
```

...which is explained by this snippet:
```js
<Icon className="bagels" name="message" /> //=> render Icon with className prop

const Icon = createComponent({
  props: () => ({
    className: 'mdi mdi-message', //=> generated by `Icon.getClassName`
  });

// and then inside of `utils#createComponent`
const resolvedProps = {
  ...baseProps(props), //=> className: mdi mdi-message
  ...props, //=> className: overwritten by bagels!!!   
};
```

### DIFF
--- a/src/Icon/__snapshots__/Icon.spec.js.snap
+++ b/src/Icon/__snapshots__/Icon.spec.js.snap
@@ -22,7 +22,7 @@ exports[`<Icon /> renders with passed class name 1`] = `
 }
 
 <i
-    class="bagels bagels mdi mdi-message re-icon c0"
+    class="bagels mdi mdi-message re-icon c0"
     name="message"
   />
 </DocumentFragment>
--- a/src/Icon/__snapshots__/Icon.spec.js.snap
+++ b/src/Icon/__snapshots__/Icon.spec.js.snap
@@ -1,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Icon /> renders 1`] = `
+<DocumentFragment>
+  .c0 {
+  color: inherit;
+  font-size: inherit;
+}
+
+<i
+    class="mdi mdi-message re-icon c0"
+    name="message"
+  />
+</DocumentFragment>
+`;
+
+exports[`<Icon /> renders with passed class name 1`] = `
+<DocumentFragment>
+  .c0 {
+  color: inherit;
+  font-size: inherit;
+}
+
+<i
+    class="bagels bagels mdi mdi-message re-icon c0"
+    name="message"
+  />
+</DocumentFragment>
+`;
--- a/src/utils.js
+++ b/src/utils.js
@@ -13,8 +13,10 @@ export const getComponentVariant = (theme, componentName, variant) => {
 
 export const getComponentStyle = componentName => themeGet(`components.${componentName}.style`);
 
-export const getComponentClassName = ({ className, theme: { classPrefix }, variant }, name) =>
-  `${className || ''} ${classPrefix}-${name} ${variant ? `${classPrefix}-${name}-${variant}` : ''}`.trim();
+export const getComponentClassName = ({ className, passedClassName, theme: { classPrefix }, variant }, name) =>
+  `${className || ''} ${passedClassName || ''} ${classPrefix}-${name} ${
+    variant ? `${classPrefix}-${name}-${variant}` : ''
+  }`.trim();
 
 export const createComponent = ({ name, tag = 'div', as, style, props: baseProps = () => ({}) }) => {
   const component = as ? styled(as) : styled[tag];
--- a/src/utils.js
+++ b/src/utils.js
@@ -13,22 +13,22 @@ export const getComponentVariant = (theme, componentName, variant) => {
 
 export const getComponentStyle = componentName => themeGet(`components.${componentName}.style`);
 
-const getComponentClassName = ({ baseClassName, theme: { classPrefix }, variant }, name) =>
-  `${baseClassName || ''} ${classPrefix}-${name} ${variant ? `${classPrefix}-${name}-${variant}` : ''}`.trim();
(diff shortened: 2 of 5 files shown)


src/Icon/__snapshots__/Icon.spec.js.snap
```diff
@@ -0,0 +1,29 @@
...
+
+exports[`<Icon /> renders with passed class name 1`] = `
+<DocumentFragment>
+  .c0 {
+  color: inherit;
+  font-size: inherit;
+}
+
+<i
+    class="bagels bagels mdi mdi-message re-icon c0"
```
erikshestopal: I have no idea why `bagels` is being duplicated here. Logging out the output of `utils#getComponentClassName` shows the correct name but when the component is rendered, this duplicate class name is added 🤷‍♂️

src/utils.js
```diff
@@ -13,8 +13,17 @@ export const getComponentVariant = (theme, componentName, variant) => {
 
 export const getComponentStyle = componentName => themeGet(`components.${componentName}.style`);
 
-export const getComponentClassName = ({ className, theme: { classPrefix }, variant }, name) =>
-  `${className || ''} ${classPrefix}-${name} ${variant ? `${classPrefix}-${name}-${variant}` : ''}`.trim();
+const getComponentClassName = flow(
```
cehsu: Could the classnames package be helpful for managing some of this logic?

erikshestopal: It definitely could be (works really awesome for this particular use case) but I'm not sure pulling in another dependency just for this function would be worth it. Thoughts?

cehsu: Oh, I assumed we were already using it. Yes, we probably wouldn't want to pull it in to use only once. There's a [dedupe](https://github.com/JedWatson/classnames#alternate-dedupe-version) feature that seems like it could be helpful.

src/Icon/__snapshots__/Icon.spec.js.snap
kylealwyn: Custom class names are indeed duplicated when rendered:

Screen Shot 2019-03-27 at 12 47 19 PM

src/utils.js
kylealwyn: I like the fp approach. Flow is great for running a series of transformations, but here we're only running one, which could be rewritten as `'hey im a string'.split(' ').filter(Boolean).join(' ')`. I also don't quite see what this is doing that `trim()` doesn't.

src/utils.js
```diff
@@ -13,8 +13,17 @@ export const getComponentVariant = (theme, componentName, variant) => {
...
+  ({ className, passedClassName, theme: { classPrefix }, variant }, name) =>
+    `${className || ''} ${passedClassName || ''} ${classPrefix}-${name} ${
```
kylealwyn: Looks like styled-components automatically applies the className passed on render undo the hood. We can fix the class duplication by removing `className` here.

Could we also change `passedClassName` to `baseClassName` to keep consistent with `baseProps`? :)

src/utils.js
```diff
@@ -13,8 +13,13 @@ export const getComponentVariant = (theme, componentName, variant) => {
 
 export const getComponentStyle = componentName => themeGet(`components.${componentName}.style`);
 
-export const getComponentClassName = ({ className, theme: { classPrefix }, variant }, name) =>
-  `${className || ''} ${classPrefix}-${name} ${variant ? `${classPrefix}-${name}-${variant}` : ''}`.trim();
+const getComponentClassName = ({ className, passedClassName, theme: { classPrefix }, variant }, name) =>
```
cehsu: I'm not sure why, but removing `className` from here fixed the duplication issue for me locally, while retaining the `className` on the element.

erikshestopal: can you get the test snapshots to pass? I've noticed the `re-box` class not being applied to `Card` when removing `className` from the props here.

cehsu: I'll check.

cehsu: Also experiencing the same failure.

src/utils.js
```diff
@@ -13,20 +13,23 @@ export const getComponentVariant = (theme, componentName, variant) => {
...
 
   return component.attrs(props => {
-    const resolvedProps = {
-      ...baseProps(props),
-      ...props,
+    const resolvedProps = { ...baseProps(props), ...props };
+
+    const finalProps = {
+      ...resolvedProps,
+      baseClassName: get(resolvedProps, 'baseClassName', props.className),
```
kylealwyn: Nice! This looks good. Sorry to be annoying but one more perf suggestion to only spread once:
```js
const baseProps = getBaseProps(props);
const finalProps = {
   ...baseProps,
   ...props,
    baseClassName: get(baseProps, 'baseClassName', props.className),
}
```

erikshestopal: no problem!

src/utils.js
cehsu: Perhaps a short-term fix could be to simply dedupe the `className` string using a set or lodash `.uniq`

src/utils.js
```diff
@@ -13,20 +13,23 @@ export const getComponentVariant = (theme, componentName, variant) => {
...
+export const createComponent = ({ name, tag = 'div', as, style, props: getBaseProps = () => ({}) }) => {
   const component = as ? styled(as) : styled[tag];
 
   return component.attrs(props => {
-    const resolvedProps = {
-      ...baseProps(props),
+    const baseProps = getBaseProps(props);
+    const finalProps = {
       ...props,
+      ...baseProps,
```
cehsu: Could we want to reverse the props and baseProps, so that props can be used to override?

PR #37, opened 2019-04-08
fix(Portal): export Portal component

erikshestopal:


src/index.js
```diff
@@ -24,6 +24,7 @@ export Input from './Form/Input';
 export Label from './Form/Label';
 export Linkify from './Linkify';
 export Modal from './Modal';
+export Portal from './Portal';
```
kylealwyn: Should come after Placeholder

PR #38, opened 2019-04-13
fix(Linkify): Skip code block formatting

cehsu:
These changes allow the `<Linkify/>` component to display text formatted as a code block (according to [gfm](https://github.github.com/gfm/)) as normal text.

image

### DIFF
--- a/src/Linkify/__snapshots__/Linkify.spec.js.snap
+++ b/src/Linkify/__snapshots__/Linkify.spec.js.snap
@@ -160,9 +160,9 @@ exports[`Linkify reformats code blocks 1`] = `
     <div>
       dear doctor,
     </div>
-    <span>
+    <div>
       I'm "5'9" and 160'
-    </span>
+    </div>
   </div>
 </DocumentFragment>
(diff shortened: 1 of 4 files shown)


cehsu:
Thanks @choochootrain, it does.

src/Linkify/__snapshots__/Linkify.spec.js.snap
```diff
@@ -107,3 +139,30 @@ exports[`Linkify escapes HTML entities 1`] = `
...
+  margin-top: 8px;
+}
+
+<div
+    class="re-linkify c0"
+  >
+    <div>
+      dear doctor,
+    </div>
+    <span>
```
cehsu: @kylealwyn are we happy removing the additional newline here? It takes 2 newlines for an indented line to be interpreted as a code block. The code renderer approach only results in one.

kylealwyn: rendering in a div gives us the top buffer that I think we're looking for

PR #39, opened 2019-04-15
Linkify proposal

kylealwyn:
@choochootrain @cehsu manipulating the source string beyond carriage return conversion seems too brittle to me. If we choose to only use a source string prop, we also drop our dependency on `renderStaticToMarkup`

### DIFF
--- a/src/Linkify/Linkify.js
+++ b/src/Linkify/Linkify.js
@@ -47,8 +47,7 @@ export default function Linkify({ source, linkStyle, renderers, ...props }) {
 }
 
 Linkify.propTypes = {
-  children: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
-  source: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
+  source: PropTypes.string,
   className: PropTypes.string,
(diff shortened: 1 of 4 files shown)


cehsu:
Is the idea to replace usages of the `children` prop with an alternative link component, such as React Router `<Link/>`?

kylealwyn:
@cehsu No, the idea is to not use the children prop at all, like react markdown

src/Linkify/Linkify.js
```diff
@@ -13,21 +12,33 @@ const Link = styled.a.attrs(() => ({
...
+  div + div {
+    margin-top: 8px;
   }
 `;
 
-export default function Linkify({ children, source, linkStyle, renderers, ...props }) {
-  let linkContent = children || source;
-  if (typeof linkContent === 'string') linkContent = linkContent.replace(/\r/g, '\n').replace(/[ \t]+/g, ' ');
+export default function Linkify({ source, linkStyle, renderers, ...props }) {
+  if (typeof source !== 'string') {
```
cehsu: Should we also update the proptypes as well?

PR #41, opened 2019-05-01
fix(Modal): Lock focus

cehsu:
These changes:
- Lock focus in modals (via `react-focus-lock`)
- Adds an example with a dropdown (to illustrate how returnFocus is handled)

image

Next steps:
- add VRT with puppeteer, and tie into CI
- consider using [react-focus-on](https://github.com/theKashey/react-focus-on) for modals as recommended by [react-focus-lock](https://github.com/theKashey/react-focus-lock#more)
- close on escape keypress

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -106,7 +106,7 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
         <Transition in={isOpen} timeout={animationDuration}>
           {state => (
             <Backdrop transitionState={state} onClick={handleBackdropClick}>
-              <FocusLock lockProps={{ style: { height: '100%' } }} disabled={!isOpen}>
+              <FocusLock lockProps={{ style: { maxHeight: '100%' } }} disabled={!isOpen}>
                 <ModalContent transitionState={state} onClick={handleContentClick} {...props}>
                   {title && <Modal.Header title={title} showClose={showClose} />}
                   {children}
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -106,7 +106,7 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
         <Transition in={isOpen} timeout={animationDuration}>
           {state => (
             <Backdrop transitionState={state} onClick={handleBackdropClick}>
-              <FocusLock disabled={!isOpen}>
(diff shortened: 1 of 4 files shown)


src/Modal/Modal.js
```diff
@@ -118,6 +121,7 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
 }
 
 Modal.propTypes = {
+  groupId: PropTypes.string.isRequired,
```
kylealwyn: What is groupId's purpose and could we get away without it? This would be a somewhat larger change given that we'd have to go and add groupId's to every existing modal.

cehsu: Thanks for pointing that out, this is zombie code from testing [a recommended approach](https://github.com/theKashey/react-focus-lock#portals) for handling portalled content, and has been removed.

cehsu:
@kylealwyn the scrolling issue is indeed new, good catch, investigating.

src/Modal/Modal.mdx
```diff
@@ -19,13 +20,21 @@ Modals are a great way to add dialogs to your site for lightboxes, user notifica
...
+  <ModalExample groupId={1} body="I am an example modal that displays example content to prove that I can actually do some things" maxWidth={300} />
 </Playground>
 
 ## Handling Long Content
 
 When modals become too long for the user’s viewport or device, they scroll independent of the page itself.
 
 <Playground>
-  <ModalExample body={<div>{new Array(50).fill(null).map((_, i) => <p key={i}>I'm really long annoying content.</p>)}</div>} />
+  <ModalExample groupdId={2} body={<div>{new Array(50).fill(null).map((_, i) => <p key={i}>I'm really long annoying content.</p>)}</div>} />
```
kylealwyn: Drop `groupId`

src/Modal/Modal.mdx
```diff
@@ -19,13 +20,21 @@ Modals are a great way to add dialogs to your site for lightboxes, user notifica
 Toggle a working modal demo by clicking the button below. It will slide up and fade in from the bottom of the page.
 
 <Playground>
-  <ModalExample body="I am an example modal that displays example content to prove that I can actually do some things" maxWidth={300} />
+  <ModalExample groupId={1} body="I am an example modal that displays example content to prove that I can actually do some things" maxWidth={300} />
```
kylealwyn: Drop `groupId`

PR #42, opened 2019-05-02
fix(Dropdown): Remove timeouts, Fix ref bug

cehsu:
These changes remove the timeout in handleMenuBlur, which was causing issues when attempting to focus on inputs while the dropdown was open, and would cause the menu not to close on blur.

Also this fixes a bug, when attempting to open the dropdown via the enter key (when menuRef.current was null).


cehsu:
Fixing some tests for escape/close and also adding one for enter, and then will open back up.

src/Dropdown/Dropdown.js
```diff
@@ -85,26 +95,25 @@ export default function Dropdown({
...
+        setBlurable(true);
       }
     }
   });
 
-  const handleMenuBlur = () => {
-    if (autoclose) {
-      // Use timeout to delay examination of activeElement until after blur/focus
-      // events have been processed.
-      setTimeout(() => {
```
cehsu: This setTimeout was causing a bug with blur when inputs were clicked. Managing whether blur is being triggered via the state now allows inputs clicks to trigger blur.

src/Dropdown/Dropdown.js
```diff
@@ -358,7 +367,7 @@ Dropdown.Footer = createComponent({
     as: 'footer',
   }),
   style: ({ theme }) => css`
-    background: ${theme.colors.grayLightest};
+    background: ${({theme}) => get(theme, 'colors.grayLightest', 'gray')};
```
cehsu: These changes are hacks to enable enzyme testing.

src/Dropdown/Dropdown.mdx
```diff
@@ -114,6 +115,7 @@ Easily display contextual overlays using custom trigger elements. Dropdown's pos
               </Dropdown>
             </Flex>
           </Button.Group>
+          <Input name="password" label="Password" />
```
cehsu: This is simply included to facilitate reviewer testing and will be removed. Ideally this would be tested via integration testing or VRT, and that's something to keep in mind moving forward.

src/Dropdown/Dropdown.js
choochootrain: do we have to hack this for all components we want to test?

kylealwyn:
@cehsu Could you elaborate on why enzyme was added? It'd be preferable to not have two testing frameworks in play simultaneously

cehsu:
@kylealwyn the initial test framework is based on direct DOM manipulation, but as noted [here](https://github.com/heydoctor/molekule/pull/42/files#diff-8f7db5ffea80e3f23edd9507af0cf020R118), I wasn't able to trigger the toggle that way in the console. Therefore enzyme which is more closely tied to the component lifecycle, was tested out, and it does the job.

src/Dropdown/Dropdown.js
cehsu: Perhaps, it's worth looking into moving forward if there's a more elegant way to handle this, assuming we turn top enzyme for more testing.

kylealwyn:
@cehsu We were using blur successfully in the tests before, so I'm not entirely convinced that's the case:
```js
    // Some issues with fireEvent.focus: https://github.com/kentcdodds/react-testing-library/issues/276#issuecomment-473392827
    renderUtils.wrapper.focus();
    renderUtils.getByTestId('dropdown-menu').blur();
```

cehsu:
@kylealwyn thanks again, I think removing the setTimeout may have affected the ability to use blur directly.

kylealwyn:
@cehsu I'd prefer we stick with react-testing-library so if you point me to the exact issue you were encountering, I'd like to take a look

cehsu:
Thanks @kylealwyn. The original (now failing) test has been restored for reference.

cehsu:
I'm restored the enzyme tests for now since they work, if that's alright. We can revisit the libraries used again.

kylealwyn:
@cehsu I'd prefer that we skip the failing test for now with manual inspection rather than adding an entirely new test framework

src/Dropdown/Dropdown.js
```diff
@@ -52,17 +52,26 @@ export default function Dropdown({
...
+    setOpen(false);
+  }
   const toggle = () => (isOpen ? close() : open());
 
   const handleTrigger = e => {
     // Allow all clicks and, for non-button elements, Enter and Space to toggle Dropdown
     // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Required_JavaScript_Features
     if (e.type === 'click' || (e.type === 'keypress' && (e.which === 13 || e.which === 32))) {
       e.stopPropagation();
+      e.preventDefault();
```
kylealwyn: flip flop with `stopPropagation` for consistency

src/Dropdown/Dropdown.js
```diff
@@ -52,17 +52,26 @@ export default function Dropdown({
 }) {
   const popperRef = useRef();
   const menuRef = useRef();
+
   const [isOpen, setOpen] = useState(false);
+  const [isBlurable, setBlurable] = useState(false);
```
kylealwyn: Could this be assigned as a ref? State should generally be used in render, otherwise it can be a ref. We also benefit from the safety of the update processing synchronously, rather than waiting for next render

cehsu: useState has been replaced. I'm not familiar with using refs in that way, could you please elaborate?

src/Dropdown/Dropdown.mdx
cehsu: Temporarily restored for reviewer manual testing.

src/Dropdown/Dropdown.js
```diff
@@ -37,6 +37,12 @@ const Trigger = createComponent({
   `,
 });
 
+let isBlurable = false;
+
+const setBlurable = value => {
+  isBlurable = value;
+}
```
erikshestopal: Can you `useRef` for this? Thinking something like
```js
const blurable = useRef(false);

const setBlurable = value => {
  blurable.current = value;
}
```

src/Dropdown/Dropdown.mdx
erikshestopal: This seems like an existing issue and may be out of scope but I have to click twice into an input field when the dropdown is open in order to get the input field to focus.

src/Dropdown/Dropdown.js
kylealwyn: Yes ^ this unfortunately won't work with multiple dropdowns because it is global.

src/Dropdown/Dropdown.mdx
kylealwyn: @erikshestopal this PR is actually addressing that issue
@cehsu experiencing the same as erik - clicking into the input does not focus the input

cehsu: That's correct, the original bug was that the dropdown would not close when an input was clicked. Now the dropdown will close when an input is clicked, and the input can be subsequently focused.

cehsu: Updated to reimplement FocusLock's `returnFocus`, which was preventing the input focus on blur.

src/Dropdown/Dropdown.js
```diff
@@ -95,16 +104,9 @@ export default function Dropdown({
...
-      // Use timeout to delay examination of activeElement until after blur/focus
-      // events have been processed.
-      setTimeout(() => {
-        const nextActiveElement = document.activeElement;
-        if (menuRef.current && !menuRef.current.contains(nextActiveElement)) {
-          close();
-        }
-      });
+  const handleMenuBlur = e => {
+    if (autoclose && menuRef.current && !menuRef.current.contains(e.relatedTarget)) {
```
cehsu: It turns out `relatedTarget` meets our needs, for determining the click target.

kylealwyn:
@cehsu Could we tackle the tests or are they still cryptically failing?

src/Dropdown/Dropdown.js
```diff
@@ -71,20 +78,22 @@ export default function Dropdown({
...
       setTimeout(() => {
-        menuRef.current.focus();
+        if (menuRef.current) menuRef.current.focus();
       });
     }
   }, [isOpen]);
 
   useKeyPress('Escape', () => {
     if (isOpen) {
+      triggerRef.current.focus();
```
cehsu: We manually implement `returnFocus` here, so that inputs can be directly focused (without a second click), which was prevented by using FocusLock's `returnFocus`.

cehsu:
@kylealwyn, good idea, the blur test re-enabled, I wasn't able to get the keypress trigger event tests to work however.

PR #46, opened 2019-07-12
feat: Checkbox color updates & Extending theme based styling to variants

jlemm45:
## Summary
- Add the ability to pass `colorOn ` & `colorOff` when using a `CheckboxGroup`
- Add the ability to extend theme based styling into variants as well
- Changes are based on issues encountered with PRD updates

## Theme styling updates

**Problem**
When extending the styles for a particular component at the theme level, I can't dig deeper into a per variant basis. This causes problems when trying to change the `hover` state for a secondary button only for example.

**Solution**
Allow for the ability to pass styles into the variant level as well. 

For example if I want to target a secondary button only I can pass this following in `theme.js`. The `style` key accepts a template string or a style object.

```
{  
  Button: {
      variants: {
        secondary: {
          style: `
            &:hover {
              background-color: tomato;
            }
          `,
        },
      },
    },
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "molekule",
   "description": "React component library built on top of styled-components and styled-system",
-  "version": "4.0.0",
+  "version": "3.5.7",
   "main": "lib/molekule.js",
   "module": "lib/molekule.es.js",
   "files": [
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "molekule",
   "description": "React component library built on top of styled-components and styled-system",
-  "version": "3.5.7",
+  "version": "4.0.0",
   "main": "lib/molekule.js",
   "module": "lib/molekule.es.js",
   "files": [
--- a/src/Form/CheckboxGroup.js
+++ b/src/Form/CheckboxGroup.js
@@ -9,7 +9,6 @@ import { createEasyInput } from './EasyInput';
 class CheckboxGroup extends Component {
   static propTypes = {
     name: PropTypes.string.isRequired,
-    color: PropTypes.string,
     horizontal: PropTypes.bool,
     value: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number])),
     choices: PropTypes.arrayOf(
@@ -20,13 +19,15 @@ class CheckboxGroup extends Component {
       })
     ).isRequired,
     onChange: PropTypes.func,
+    colorOn: PropTypes.string,
+    colorOff: PropTypes.string,
   };
 
   static defaultProps = {
-    defaultValue: [],
-    color: 'primary',
     horizontal: false,
     onChange() {},
+    colorOn: 'primary',
+    colorOff: 'grayMid',
   };
 
   state = {
@@ -60,7 +61,7 @@ class CheckboxGroup extends Component {
   };
 
   render() {
-    const { name, color, choices, error, horizontal } = this.props;
+    const { name, choices, error, horizontal, colorOn, colorOff } = this.props;
     const { selected } = this.state;
 
     return (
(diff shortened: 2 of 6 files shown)


src/utils.js
```diff
@@ -4,15 +4,19 @@ import styled from 'styled-components';
...
   });
+
   if (!config) throw new Error(`Molekule: "${variant}" variant not found in theme...`);
   return config;
 };
 
 export const getComponentStyle = componentName => themeGet(`components.${componentName}.style`);
 
+export const getComponentVariantStyles = (componentName, variant) =>
+  themeGet(`components.${componentName}.variants.${variant}.style`);
```
erikshestopal: This will return a function instead of a value because `themeGet` returns a function that accepts `theme` as an argument.

src/utils.js
```diff
@@ -4,15 +4,19 @@ import styled from 'styled-components';
 export const themeGet = (lookup, fallback) => ({ theme } = {}) => get(theme, lookup, fallback);
 
 export const getComponentVariant = (theme, componentName, variant) => {
-  const config = themeGet(`components.${componentName}.variants.${variant}`, theme.variants[componentName][variant])({
+  const config = themeGet(`variants.${componentName}.${variant}`, theme.variants[componentName][variant])({
```
cehsu: We are now using the same thing for the lookup and the fallback?

src/utils.js
cehsu: It looks like we want to support two conventions: `components.${componentName}.variants.${variant}` and also `variants.${componentName}.${variant}` (similar to `getComponentVariant`)?

jlemm45: I believe I addressed this here https://github.com/sappira-inc/molekule/blob/0656ca13a18efe5b70ebbced48f3848572fe321f/src/utils.js#L20 LMK

src/utils.js
jlemm45: @cehsu I bumped the major version

src/utils.js
jlemm45: This seems to be the way it has been used here https://github.com/sappira-inc/molekule/blob/0656ca13a18efe5b70ebbced48f3848572fe321f/src/utils.js#L15 ? LMK

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "molekule",
   "description": "React component library built on top of styled-components and styled-system",
-  "version": "3.5.7",
+  "version": "4.0.0",
```
kylealwyn: We version bump on release post-merge

jlemm45:
@kylealwyn Per @cehsu's comments I've already included that pattern you mention in tandem with the pattern above so you can do the following.
```
{
  Button: {
      variants: {
        secondary: {
          style: `
            &:hover {
              background-color: tomato;
            }
          `,
        },
      },
  }
}
```
or what you mention above
```
{
  primary: {
    backgroundColor: colors.primary,
    fontColor: 'white',
    style: css`
	  &:hover {
         background: coral;
      },
  },
}
```

kylealwyn:
@jlemm45 @cehsu I don't see why we would need or want to support both methods of styling variants

jlemm45:
@kylealwyn @cehsu Happy to remove the first pattern if no one sees a need.

kylealwyn:
@jlemm45 Lets remove

kylealwyn:
@heydoctor/engineering Given this is already going going to be a breaking change, albeit a minor, one file break, I propose we convert component style overrides to the same pattern as variants:
```js
const theme = {
	...,
    variants: {
        Button: { ... },
    },
    styles: {
        Button: `
          border-radius: 42px;
        `
     }
}

```

src/Form/CheckboxGroup.js
```diff
@@ -77,7 +78,8 @@ class CheckboxGroup extends Component {
                   id={key}
                   name={key}
                   label={choice.label}
-                  color={color}
+                  colorOn={colorOn}
+                  colorOff={colorOff}
```
cehsu: 🎉

jlemm45:
@cehsu Are referring to testing the new util functions in `utils.js`? Or maybe adding a test to cover applying the different styling patterns to a component?

jlemm45:
@cehsu I added some tests. LMK if you were thinking of something different. @kylealwyn Any further thoughts?

PR #47, opened 2019-07-24
feat(Dropdown): Add support for string widths

kylealwyn:
## Summary
This sets the default width as auto (150 is too arbitrary) and adds general support for string widths, such as 1rem.

## Why?
To account for dynamically sized dropdowns
https://app.asana.com/0/1131771498685767/1132620207943367


kylealwyn:
@erikshestopal Added tests. Can you give another look?

kylealwyn:
@erikshestopal It's for SSR, where Element is undefined. See https://github.com/heydoctor/molekule/pull/43

PR #49, opened 2019-07-29
feat(Input): `isFocused` state and icon support

mathewmorris:
## Summary
- Adds `isFocused` state to label
- Support for adding icons to input
- Adds `SearchInput` component

## Why?
- To have the ability to style labels when input is focused
- Input styled updates
- Needed an input that was slightly different, but also wasn't going to interfere with original `Input` component

https://app.asana.com/0/1129589982637900/1129773484722494


src/Form/Input.js
```diff
@@ -271,12 +306,14 @@ class Input extends Component {
         <InputContainer styles={rest.styles}>
           {floating && Label}
 
-          {multiline ? <StyledTextArea {...inputProps} /> : <StyledInput {...inputProps} />}
+          {icon && <StyledIcon size={iconSize} color={iconColor} name={icon} />}
+
+          {multiline ? <StyledTextArea {...inputProps} /> : <StyledInput hasIcon={icon} {...inputProps} />}
```
cehsu: Should we pass in a boolean here?
```suggestion
          {multiline ? <StyledTextArea {...inputProps} /> : <StyledInput hasIcon={!!icon} {...inputProps} />}
```

mathewmorris: Ah yes. That makes more sense.

src/Form/Input.js
```diff
@@ -271,12 +325,18 @@ class Input extends Component {
         <InputContainer styles={rest.styles}>
           {floating && Label}
 
-          {multiline ? <StyledTextArea {...inputProps} /> : <StyledInput {...inputProps} />}
+          {icon && <StyledIcon size={iconSize} color={iconColor} name={icon} />}
+
+          {multiline ? <StyledTextArea {...inputProps} /> : <StyledInput hasIcon={!!icon} {...inputProps} />}
+
+          {isClearable && stateValue && (
+            <CloseIcon name="close-circle" color="greyDarker" size={18} onClick={this.handleClear} />
```
kylealwyn: What do you think about making this more configurable? We have a small eye on our mobile app that toggles password visibility - this may be a good way to do this extensibly on the web. The tradeoff will add a little more boilerplate code for clearability.

```jsx
<Input
  leftIcon="security-lock"
  leftIconSize={10}
  leftIconColor={14}
  onLeftIconClick={}
  rightIcon="eye-closed"
  rightIconSize
  rightIconColor
  onRightIconClick={}
/>
```

mathewmorris: I like it. We are also adding the eye to the web app as well, which will be helpful there too. 👍

src/Form/Input.js
```diff
@@ -112,6 +149,8 @@ class Input extends Component {
     onBlur() {},
     onChange() {},
     floating: false,
+    iconColor: 'greyDark',
+    iconSize: 18,
```
kylealwyn: need to update default props

src/Form/Input.js
```diff
@@ -112,6 +149,8 @@ class Input extends Component {
     onBlur() {},
     onChange() {},
     floating: false,
+    iconColor: 'greyDark',
```
kylealwyn: we'll need to update molekule default theme from gray to grey

src/Form/Input.js
```diff
@@ -46,7 +47,17 @@ const StyledInput = createComponent({
 
     ${isFloating &&
       css`
-        padding-bottom: 0;
+        padding-bottom: 0px;
+      `};
+
+    ${hasLeftIcon &&
+      css`
+        padding-left: 28px;
```
kylealwyn: these should probably be dynamic. `leftIconSize + 10`??

src/Form/Input.js
```diff
@@ -238,28 +300,50 @@ class Input extends Component {
...
       onFocus: this.onFocus,
       onBlur: this.onBlur,
-      style: multiline ? { ...style, height: this.state.height } : style,
+      style: multiline ? { ...style, height } : style,
       placeholder,
       isFloatable: floating,
       isFloating,
+      error,
+      leftIcon,
+      leftIconSize,
```
kylealwyn: I don't think these need to be passed to the actual input element

mathewmorris: We need a way to pass the `iconSize` to add the appropriate padding to either side of the input if either `icon` is present. 

I suppose we could turn the `leftIcon` into `hasLeftIcon: !!leftIcon` to reduce the explicit `hasLeftIcon={!!leftIcon} hasRightIcon={!!rightIcon}` passed to the input.

That's still passing them through though. :/

src/Form/SearchInput.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    const { input } = renderInput({ value });
+
+    expect(input.value).toEqual('Search item');
+  });
+
+  test('updates internally when value prop changes', () => {
+    const { input, rerender } = renderInput({ value: 'Searching' });
+    rerender(
+      <ThemeProvider>
+        <SearchInput placeholder="Search" />
```
kylealwyn: This isn't doing what the test condition is describing - I think we'll need to add an updated value prop here?

src/Form/SearchInput.spec.js
```diff
@@ -0,0 +1,36 @@
...
+      </ThemeProvider>
+    );
+    expect(input.value).toEqual('Searching');
+  });
```
kylealwyn: Can we add a test case for clearing the value? So we find the icon, click it, and expect the input value to be empty?

src/Form/Input.js
kylealwyn: Makes sense - a little verbose but that seems fine

src/Form/SearchInput.spec.js
mathewmorris: I'm working on this one right now :)

src/Form/SearchInput.spec.js
mathewmorris: I removed this test since there is no reformatting `onChange` like there is in the `DateInput` and `PhoneInput`.

mathewmorris:
That puts us at 100% coverage on `SearchInput`. At least according to jest.

src/Form/SearchInput.spec.js
mathewmorris: Though, this could be covered with a `useEffect` hook. Not really sure how this would be beneficial in the search input since it never needs to reformat or be rehydrated from any state.

src/Form/Input.mdx
```diff
@@ -49,57 +50,79 @@ Form input component. Different sizes are available.
...
 ### Date Input
-<DateInput placeholder="Date of Birth" label="Date of Birth" initialValue="04/01/2001" value="2001-04-01" />
+<Playground>
+  <DateInput placeholder="Date of Birth" label="Date of Birth" initialValue="04/01/2001" value="2001-04-01" />
+</Playground>
+
+### Search Input
+<Playground>
+  <SearchInput placeholder="Search" label="Search" />
+</Playground>
```
cehsu: I'm not seeing this example locally, is that just me?

mathewmorris: You might need to run `npm run docs` again? I had the issue early on using `yarn`.

src/Form/Input.js
```diff
@@ -221,6 +278,11 @@ class Input extends Component {
     this.ref.current.blur();
   }
 
+  handleClear = () => {
+    this.setState({ value: '' });
+    this.props.onChange(this.inputRef.current.name, '');
```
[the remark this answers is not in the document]
mathewmorris: I actually just removed that `handleClear` method. It's not needed in the Input since that's specific to another component. I believe this duplication is taken care of as well with this removal.

src/Form/Input.js
```diff
@@ -65,6 +85,29 @@ const AutogrowShadow = createComponent({
...
+  style: css`
+    position: absolute;
+    left: 8px;
+    top: 50%;
+    transform: translateY(-50%);
+  `,
+});
+
+const RightIcon = createComponent({
+  name: 'LeftIcon',
```
jlemm45: Should this be `RightIcon`?

src/Form/Input.js
```diff
@@ -65,6 +85,29 @@ const AutogrowShadow = createComponent({
...
+  as: Icon,
+  style: css`
+    position: absolute;
+    right: 8px;
+    top: 50%;
+    transform: translateY(-50%);
+    cursor: pointer;
```
jlemm45: Observation - `RightIcon` has a `cursor:pointer` but not `LeftIcon`. Is this intended?

src/Form/SearchInput.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Search
+      forwardedRef={inputRef}
+      leftIcon="magnify"
+      rightIcon={currentValue && 'close-circle'}
+      onRightIconClick={clearValue}
+      onChange={handleChange}
+      value={currentValue}
+      size="sm"
```
jlemm45: I think in other components `md` seems to be the default. Just wondering if this should be the case?

src/Form/SearchInput.js
```diff
@@ -0,0 +1,55 @@
+import React, { useRef, useState } from 'react';
+import { css } from 'styled-components';
+import { createComponent } from '../utils';
+import { createEasyInput } from './EasyInput';
+import Input from './Input';
+
+const Search = createComponent({
```
jlemm45: Wondering if we can get `prop-types` with defaults for this component? Makes the docz a little nicer with a Props Table.

src/utils.js
```diff
@@ -90,3 +90,6 @@ export const getNextCursorPosition = (cursorPos, oldValue, newValue) => {
 
   return Math.max(nextPosition, 0);
 };
+
+export const getOneSizeSmaller = (collection, size) =>
```
jlemm45: Maybe a test for this util method?

src/Form/Input.js
mathewmorris: Yup.

src/Form/SearchInput.js
mathewmorris: I used `sm` since it follows the design, but, we could leave it at default `md` and then use `sm` in app. What do you think?

src/Form/Input.js
mathewmorris: Well, same thing goes here as for the search input default size. We could leave as default no cursor changes, and allow that to be determined in app.

src/utils.js
mathewmorris: Ah yes. Sounds like a good idea.

src/Form/Input.js
mathewmorris: Going to leave the cursor style up to user.

src/Form/SearchInput.js
mathewmorris: Going to leave input size up to user.

src/Form/Input.js
```diff
@@ -33,11 +43,11 @@ const StyledInput = createComponent({
     &:hover,
     &:focus,
     &:active {
-      border-color: ${theme.colors.grayMid};
+      border-color: ${theme.colors.grey};
```
erikshestopal: The Figma spec calls for a blue border for active inputs. Wanted to confirm that this change was intentional.

src/Form/Input.js
```diff
@@ -65,6 +85,38 @@ const AutogrowShadow = createComponent({
...
+  style: ({ onClick }) => css`
+    position: absolute;
+    right: 8px;
+    top: 50%;
+    transform: translateY(-50%);
+
+    ${onClick &&
+      css`
+        cursor: pointer;
+      `}
```
erikshestopal: Styles for both left and right icons are __almost__ identical. Can we consolidate somehow?

src/Form/Input.js
```diff
@@ -112,6 +172,12 @@ class Input extends Component {
     onBlur() {},
     onChange() {},
     floating: false,
+    leftIconSize: 16,
+    leftIconColor: 'greyDarkest',
+    onLeftIconClick: null,
```
erikshestopal: Would it make sense to have empty functions as default values? 
```suggestion
    onLeftIconClick: () => {},
```

src/Form/SearchInput.js
```diff
@@ -0,0 +1,54 @@
...
+  const handleChange = (name, newValue, event) => {
+    setValue(newValue);
+
+    if (onChange) {
+      onChange(name, newValue, event);
+    }
+  };
+
+  const clearValue = () => {
+    handleChange('search', '');
```
erikshestopal: Where is `'search'` coming from?

src/Form/Input.js
mathewmorris: I'm using the null value to check if there is a function or not for the icons. If so, the icon will have `cursor: pointer` added to the styles. Otherwise, the cursor will not change.

src/Form/Input.js
mathewmorris: You're right. I'll update.

src/Form/Input.js
mathewmorris: Yeah, let me give it a shot 👍

src/Form/SearchInput.js
mathewmorris: I updated to use name coming from props :)

src/Form/Input.js
```diff
@@ -46,7 +60,19 @@ const StyledInput = createComponent({
 
     ${isFloating &&
       css`
-        padding-bottom: 0;
+        line-height: ${getOneSizeLarger(theme.fontSizes, size)}px;
```
erikshestopal: It seems to me that these utilities (`getOneSizeLarger`/`getOneSizeSmaller`) are a bit confusing and introducing unnecessary complexity. Have you considered using native css `calc()` to achieve a similar result here and elsewhere these utils are used?

```css
css`
  line-height: calc(${theme.fontSizes[size]}px * 1.2);
  padding-top: calc(${theme.fontSizes[size]}px * 1.2);
`
```

src/Form/SearchInput.js
```diff
@@ -0,0 +1,62 @@
+import React, { useRef, useState } from 'react';
```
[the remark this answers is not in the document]
mathewmorris: I don't see any problem with that. 👍

src/Form/Input.js
mathewmorris: That was my original solution, but since Jay had made very specific outlines to what the different forms of spacing/sizing needed to be, I created these utils in order to achieve that accurately instead of using `calc`. But, if Jay has no problem with it, I'm game.

mathewmorris: I'll switch to `calc` and have him review.

mathewmorris: The main place I see a problem is when these utils are changing font sizes.

src/Form/Input.js
```diff
@@ -46,7 +60,19 @@ const StyledInput = createComponent({
...
+      `};
+
+    ${hasLeftIcon &&
+      css`
+        padding-left: ${leftIconSize + 12}px;
+      `};
+
+    ${hasRightIcon &&
+      css`
+        padding-right: ${rightIconSize + 32}px;
```
cehsu: An alternative to passing down the icon props as already discussed [here](https://github.com/heydoctor/molekule/pull/49#discussion_r309821111), would be to pass down the icon component as a `rightIcon` or `leftIcon` prop (for example): `rightIcon={<Icon name="..." color="..." size={#}/>}` and its props could be accessed via:
```
${rightIcon &&
  css`
    padding-right: ${rightIcon.props.size + 32}px;
...
```

src/theme.js
```diff
@@ -95,28 +95,36 @@ export default (overrides = {}) => {
       backgroundColor: colors.blueLight,
       fontColor: colors.blueDark,
     },
-    gray: {
-      backgroundColor: colors.grayLight,
-      fontColor: colors.grayDark,
+    grey: {
```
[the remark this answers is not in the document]
mathewmorris: This is the spelling that was used in Figma, we went with it in the Color Updates.

src/theme.js
```diff
@@ -95,28 +95,36 @@ export default (overrides = {}) => {
...
-    xs: 8,
-    sm: 10,
-    md: 12,
-    lg: 14,
-    xl: 16,
+    xs: 10,
+    sm: 12,
+    md: 14,
+    lg: 16,
+    xl: 18,
```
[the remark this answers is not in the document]
mathewmorris: I could see this being beneficial as another PR, since this will be used by a few different component updates that are already PRs.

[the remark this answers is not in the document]
mathewmorris: But I think we could even cut a branch for all of these current PRs and have them as one major bump.

src/Form/Input.js
mathewmorris: @cehsu I've made some changes to try this example you've given. Let me know what you think 👍

mathewmorris: I'm also curious about some other possible ways, what do you think:
`<Input leftIcon="alert-circle" styles={{ leftIcon: { color: 'primary' } }} />`
or
`<Input leftIcon={{ name: 'alert-circle', size: 16, color: 'primary' }} />`

src/Form/Input.js
```diff
@@ -17,27 +17,31 @@ const InputContainer = createComponent({
...
+    &:focus {
+      border-color: ${theme.colors.primary};
+    }
+
     ::placeholder {
-      color: ${theme.colors.greyDark};
+      color: ${theme.colors.grey};
     }
 
     &[disabled] {
```
erikshestopal: The disabled UI in FF is inconsistent with Chrome/Safari. Can we look into making FF have a similar look as other browsers? 
Screen Shot 2019-08-19 at 10 25 52 AM

src/Form/Select.js
```diff
@@ -121,6 +121,7 @@ class Select extends Component {
         <SelectContainer value={value} size={size}>
           <SelectInput
             {...props}
+            size={size}
```
erikshestopal: Another FF quirk: select options are grayed until you actually select one. 

Screen Shot 2019-08-19 at 10 35 18 AM

src/Form/Input.js
cehsu: My thinking for removing the hasLeftIcon, and the hasRightIcon styles is that the padding amounts might be non-standard (i.e. not always 12px). Instead users could specify subcomponent styles via a styles prop (as mentioned in the overridable styles section):
```
<Input styles={{Input: { ... }}}/>
```

If we pass the Icon components in, rather than the icon props we might be able to simplify this component (so it doesn't have to manage the Icon implementation), and also make it more flexible. How does that sound @mathewmorris?

mathewmorris: @cehsu That sounds great. I think keeping the Input out of the loop of Icon component is a good thing. Just render what I give you. 👍

src/theme.js
```diff
@@ -122,19 +122,27 @@ export default (overrides = {}) => {
...
+  };
+
+  const inputHeights = {
+    xs: 32,
+    sm: 40,
+    md: 48,
```
kylealwyn: Curious why the heights are diverging - if I put a default input next to a default button, I'd expect them to be the same height.

src/Form/Input.js
cehsu: Sounds good, the approach of having the input handle the icon implementation also has precedent, but would probably require additional variables such as how [Semantic UI](https://react.semantic-ui.com/elements/input/#variations-icon) handles this via an iconMargin variable:
```
.ui.icon.input > input {
  padding-right: @iconMargin !important;
}
```
For now, an approach like [react-select](https://react-select.com/home#custom-styles), where custom styles get passed in seems like a closer delta.

src/Form/Input.js
```diff
@@ -28,29 +29,69 @@ const StyledInput = createComponent({
...
       css`
-        padding-bottom: 0;
+        line-height: 14px;
+        padding-top: 14px;
+        padding-bottom: 0px;
+      `};
+
+    ${leftIcon &&
+      css`
+        padding-left: ${(leftIconProps && leftIconProps.size) || 16 + 12}px;
```
cehsu: Rather than hard coding this padding value, how about using the icon's margin style prop. That way different layouts can be supported without requiring overwriting the input styles.
Screen Shot 2019-08-21 at 2 11 04 PM

mathewmorris: Using the icon's margin wouldn't affect the input though, since it is absolutely positioned. I believe that line isn't working as expected.

cehsu: I mean, using the icon's margin to determine the input's padding, sorry for being unclear.

mathewmorris: The user should be able to determine that gap in a few ways, here's one within the input itself:
image

They should also be able to change that in the `theme.js` file as well since we also have `InputLeftIcon` & `InputRightIcon`.

PR #50, opened 2019-07-31
fix(Input): Display floating input label if value is not undefined

cehsu:
This change ensures that floating labels appear for zero values.

### Bug:
1. Blur out of an input with a floating label, and a value of 0 (can be tested via `value={0}`).
3. We expect the label to remain, but it doesn't: 
Screen Shot 2019-07-31 at 1 45 31 PM

### Cause:
1. Zero values were not being accounted for in checking whether to display floating labels. Checking for zero values allows the label to be displayed:
Screen Shot 2019-07-31 at 1 58 38 PM

### DIFF
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -241,7 +241,7 @@ class Input extends Component {
       ...rest
     } = this.props;
 
-    const isFloating = (floating && (value && value.length > 0) || (this.state.value || this.state.value === 0));
+    const isFloating = floating && ((value && value.length > 0) || (this.state.value || this.state.value === 0));
 
     const inputProps = {
       ...rest,
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -241,7 +241,7 @@ class Input extends Component {
       ...rest
     } = this.props;
 
-    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value !== undefined);
+    const isFloating = (floating && (value && value.length > 0) || (this.state.value || this.state.value === 0));
 
     const inputProps = {
       ...rest,
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -241,7 +241,7 @@ class Input extends Component {
       ...rest
     } = this.props;
 
-    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value);
+    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value !== undefined);
 
     const inputProps = {
       ...rest,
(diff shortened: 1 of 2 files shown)


src/Form/Input.js
```diff
@@ -241,7 +241,7 @@ class Input extends Component {
       ...rest
     } = this.props;
 
-    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value);
+    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value !== undefined);
```
erikshestopal: Should we account for `null` as well?

src/Form/Label.js
```diff
@@ -9,7 +9,6 @@ const Label = createComponent({
     transition: 250ms;
     font-weight: 500;
     margin: 0 0 4px 4px;
-    z-index: 10;
```
cehsu: @schlegz noted that dropdowns and floating input labels have competing z-indices:

Screen Shot 2019-07-31 at 12 24 45 PM

Removing the z-index still achieves the same visual appearance, and resolves the conflict.
image


Are there any use cases for this z-index? Do we still want to keep it, but just bump it down to something less than 10?

src/Form/Input.js
cehsu: It couldn't hurt, what's the use case for `null`?

cehsu: Rather than check all the falsy cases, let's just stick with the truthy case, and add a check for zero, since it's the only falsy case we want to include.

src/Form/Input.js
```diff
@@ -241,7 +241,7 @@ class Input extends Component {
       ...rest
     } = this.props;
 
-    const isFloating = (floating && value && value.length > 0) || (floating && this.state.value);
+    const isFloating = floating && ((value && value.length > 0) || (this.state.value || this.state.value === 0));
```
schlegz: A bit confused on why there is a comparison against both the `props.value` and `state.value`.  Shouldn't we take in the `props.value` and set `state.value` during initialization (or props updates) and only refer to the `state.value` throughout the component, since that will always hold the current input value?

kylealwyn: I agree here. `state.value` is kept in sync with `props.value` via `getDerivedStateFromProps`. I think the condition can be summed as ```floating && value != undefined && `${value}`.length > 0```. Note we're not doing strict equality here so we capture both null and undefined values since `null == undefined`. The template string will coalesce a number into a string so we can check the length property.

Looking at you https://github.com/tc39/proposal-nullish-coalescing to make this easier

PR #51, opened 2019-08-02
feat(Button): Icon support and default style updates

jlemm45:
### Update `molekule` default styles

**Summary**
Several updates to align with new HeyDoctor brand including:
- Default colors update
- Button refresh
- Ability to pass button icon prop to display an icon in the button
- Updating docs to use default HeyDoctor fonts
- Updates `rollup` (had an issue compiling on current version)
- Update tests
- Minor updates to components with new colors

**Risk factor**
Some of the color updates could introduce a breaking change so I can consider just deprecating after we evaluate in the apps.

**New button page**

### DIFF
--- a/docs/wrapper.js
+++ b/docs/wrapper.js
@@ -1,22 +1,37 @@
-import React, { Component } from 'react';
-import { createGlobalStyle } from 'styled-components';
+import React from 'react';
+import { createGlobalStyle, css } from 'styled-components';
 import { ThemeProvider } from '../src';
 
-const GlobalStyle = createGlobalStyle `
-  * {
-    box-sizing: border-box;
-  }
-`
+const GlobalStyle = createGlobalStyle`
+  ${({ theme }) => css`
+    ${theme.fonts.map(
+      font => css`
+                @font-face {f
+                  font-family: "${font.name}";
+                  font-weight: ${font.weight || 400};
+                  font-style: normal;
+                  src: url(${font.url}) format("${font.format || 'woff2'}");
+                }
+              `
+    )}
 
-class App extends Component {
-  render() {
-    return (
-      <>
-        <GlobalStyle />
-        <ThemeProvider>{this.props.children}</ThemeProvider>
-      </>
-    );
-  }
-}
+    * {
+      font-family: "${theme.typography.bodyFontFamily}", sans-serif !important;
+    }
 
-export default App;
+    h1 {
+      font-family: "${theme.typography.headerFontFamily}", serif !important;
+      font-size: 38px;
+      color: ${theme.colors.black};
+    }
+  `}
+`;
+
+export default ({ children }) => (
+  <ThemeProvider>
+    <>
+      <GlobalStyle />
+      {children}
+    </>
+  </ThemeProvider>
+);
--- a/docs/wrapper.js
+++ b/docs/wrapper.js
@@ -1,30 +1,23 @@
 import React from 'react';
-import { createGlobalStyle, css } from 'styled-components';
+import { createGlobalStyle } from 'styled-components';
 import { ThemeProvider } from '../src';
 
 const GlobalStyle = createGlobalStyle`
-  ${({ theme }) => css`
-    ${theme.fonts.map(
-      font => css`
-        @font-face {
-          font-family: "${font.name}";
-          font-weight: ${font.weight || 500};
-          font-style: normal;
-          src: url(${font.url}) format("${font.format || 'woff2'}");
-        }
-      `
-    )}
+    @font-face {
+      font-family: "Tiempos";
+      font-weight: 500;
+      font-style: normal;
+      src: url(//cdn.heydoctor.com/fonts/TiemposHeadlineWeb-Semibold.woff2) format('woff2');
+    }
 
     * {
-      font-family: "${theme.typography.bodyFontFamily}", sans-serif !important;
+      font-family: "Avenir", sans-serif !important;
     }
 
     h1 {
-      font-family: "${theme.typography.headerFontFamily}", serif !important;
+      font-family: "Tiempos", serif !important;
       font-size: 38px;
-      color: ${theme.colors.black};
     }
-  `}
 `;
 
 export default ({ children }) => (
--- a/docs/wrapper.js
+++ b/docs/wrapper.js
@@ -6,13 +6,13 @@ const GlobalStyle = createGlobalStyle`
   ${({ theme }) => css`
     ${theme.fonts.map(
       font => css`
-                @font-face {f
-                  font-family: "${font.name}";
-                  font-weight: ${font.weight || 500};
-                  font-style: normal;
-                  src: url(${font.url}) format("${font.format || 'woff2'}");
-                }
-              `
+        @font-face {
+          font-family: "${font.name}";
+          font-weight: ${font.weight || 500};
+          font-style: normal;
+          src: url(${font.url}) format("${font.format || 'woff2'}");
+        }
+      `
     )}
 
     * {
--- a/docs/wrapper.js
+++ b/docs/wrapper.js
@@ -8,7 +8,7 @@ const GlobalStyle = createGlobalStyle`
       font => css`
                 @font-face {f
                   font-family: "${font.name}";
-                  font-weight: ${font.weight || 400};
+                  font-weight: ${font.weight || 500};
                   font-style: normal;
                   src: url(${font.url}) format("${font.format || 'woff2'}");
                 }
--- a/src/Badge/__snapshots__/Badge.spec.js.snap
+++ b/src/Badge/__snapshots__/Badge.spec.js.snap
@@ -3,13 +3,11 @@
 exports[`Badge 1`] = `
 <DocumentFragment>
   .c0 {
-  padding: 4px 8px;
-  font-size: 12px;
+  padding: 4.666666666666667px 9.333333333333334px;
+  font-size: 14px;
   font-family: inherit;
   font-weight: bold;
-  border-radius: 12px;
-  background: #C8E8FF;
-  color: #006DC1;
+  border-radius: 14px;
 }
 
 <span
--- a/src/Badge/__snapshots__/Badge.spec.js.snap
+++ b/src/Badge/__snapshots__/Badge.spec.js.snap
@@ -7,7 +7,7 @@ exports[`Badge 1`] = `
   font-size: 14px;
   font-family: inherit;
   font-weight: bold;
-  border-radius: 12px;
+  border-radius: 14px;
   background: #CADCFF;
   color: #0958F3;
 }
--- a/src/Badge/__snapshots__/Badge.spec.js.snap
+++ b/src/Badge/__snapshots__/Badge.spec.js.snap
@@ -3,7 +3,7 @@
 exports[`Badge 1`] = `
 <DocumentFragment>
   .c0 {
-  padding: 4.666666666666667px 9.333333333333334px;
+  padding: 5px 10px;
   font-size: 14px;
   font-family: inherit;
   font-weight: bold;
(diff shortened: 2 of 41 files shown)


docs/wrapper.js
```diff
@@ -1,22 +1,37 @@
...
-    box-sizing: border-box;
-  }
-`
+const GlobalStyle = createGlobalStyle`
+  ${({ theme }) => css`
+    ${theme.fonts.map(
+      font => css`
+                @font-face {f
+                  font-family: "${font.name}";
+                  font-weight: ${font.weight || 400};
```
mathewmorris: Since normal font-weight is now 500, we should default to it;

src/Card/__snapshots__/Card.spec.js.snap
```diff
@@ -5,13 +5,13 @@ exports[`Card 1`] = `
...
   box-shadow: 0 3px 6px hsla(0,0%,60%,.1),0 3px 6px hsla(0,0%,60%,.15),0 -1px 2px hsla(0,0%,60%,.02);
-  border-radius: 4px;
+  border-radius: 8px;
   overflow: hidden;
 }
 
 .c1 {
   padding: 1rem;
-  border-bottom: 1px solid #DEE0E4;
+  border-bottom: 1px solid;
```
mathewmorris: Looks like color is being dropped from border. Is this intended?

src/theme.js
```diff
@@ -95,27 +184,27 @@ export default (overrides = {}) => {
...
     },
   };
 
   const alertVariants = badgeVariants;
 
   const heights = {
     xs: 28,
-    sm: 32,
-    md: 36,
+    sm: 24,
```
mathewmorris: xs being 28 and sm doesn't really make sense. Should we switch these around?

src/theme.js
```diff
@@ -95,27 +184,27 @@ export default (overrides = {}) => {
...
+    md: 32,
     lg: 40,
-    xl: 44,
+    xl: 48,
   };
 
   const fontSizes = {
     xs: 8,
     sm: 10,
```
mathewmorris: I think 12 makes more sense here.

```suggestion
    sm: 12,
```

src/Card/__snapshots__/Card.spec.js.snap
mathewmorris: I think it may have something to do with `gray` => `grey`

src/Button/Button.js
```diff
@@ -71,34 +71,59 @@ const StyledButton = createComponent({
...
+      border-width: 1px;
       transition: 175ms;
+      white-space: nowrap;
 
       ${loading && loadingCss({ height, fontColor, outline, backgroundColor })};
 
       &:hover {
-        background: ${outline ? 'transparent' : lighten(0.05, backgroundColor)};
-        border-color: ${lighten(0.05, backgroundColor)};
+        background: ${outline || transparent ? 'transparent !important' : theme.colors.primaryLight};
```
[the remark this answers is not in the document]
erikshestopal: [QA]
Outline button not getting a background fill on hover according to Figma specs.

src/Badge/__snapshots__/Badge.spec.js.snap
```diff
@@ -3,13 +3,13 @@
 exports[`Badge 1`] = `
 <DocumentFragment>
   .c0 {
-  padding: 4px 8px;
-  font-size: 12px;
+  padding: 4.666666666666667px 9.333333333333334px;
```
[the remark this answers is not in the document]
mathewmorris: I saw this too. The styles for badges will need to be updated to cleaner numbers. Essentially it's taking the theme sizes and dividing by 3 and 1.5 respectively. Which create weird numbers with the new sizes.

erikshestopal:
[QA]:
An outline button with `variant="grey"` renders text that is white. 
Screen Shot 2019-08-09 at 5 44 17 PM

Button with no variant and `transparent=true` renders invisible
Screen Shot 2019-08-09 at 5 59 17 PM

Button in group with loading prop looks off 
btn-loading-group

src/Button/Button.mdx
```diff
@@ -5,42 +5,73 @@ menu: Core
...
 
 ## Basic usage
 
 <Playground>
-  <Button.Group>
+  <Flex justifyContent="space-around">
     <Button>Primary Button</Button>
-    <Button outline>Outline Button</Button>
-  </Button.Group>
+    <Button outline>Primary Button</Button>
```
cehsu: nit: Outline Button
```suggestion
    <Button outline>Outline Button</Button>
```

src/Button/Button.mdx
```diff
@@ -83,16 +118,22 @@ Use the `size` prop to change the size of your button.
...
 
 Add the `loading` prop for loading button state style.
 
 <Playground>
-  <Button.Group>
+  <Flex justifyContent="space-around">
     <Button loading>I'm loading</Button>
     <Button loading outline>I'm loading</Button>
-  </Button.Group>
+  </Flex>
```
cehsu: Why were button groups replaced with flex? Should Button.Group be updated to accommodate this use case, if it is common?

src/Form/Switch.mdx
```diff
@@ -4,15 +4,15 @@ name: Switch
 ---
 
 import { Playground, PropsTable } from 'docz'
-import Switch from './Switch'
+import Switch, { Switch as SwitchClass } from './Switch'
 
 # Switch
 
-Form Switch component. Different s are available.
+Form Switch component. Different styles are available.
```
cehsu: 🙏

src/Form/Switch.js
```diff
@@ -61,7 +61,7 @@ class Switch extends React.Component {
   };
 
   static defaultProps = {
-    defaultValue: false,
+    value: false,
```
cehsu: 👍

src/theme.js
```diff
@@ -45,80 +53,152 @@ export default (overrides = {}) => {
...
+      },
     },
-    success: {
-      backgroundColor: colors.green,
-      fontColor: 'white',
+    secondary: {
+      backgroundColor: colors.white,
+      fontColor: colors.primary,
+      borderColor: colors.primary,
+      textColor: colors.primary,
```
cehsu: font color and text color? can these be consolidated?

src/Button/Button.js
```diff
@@ -71,34 +81,104 @@ const StyledButton = createComponent({
...
+  `}
+`;
+
+const Button = React.forwardRef((props, ref) => {
+  const hasText = !!props.children;
+
+  return (
+    <StyledButton {...props} ref={ref} hasText={hasText}>
+      {props.icon ? (
+        <Flex alignItems="center">
```
cehsu: Is there any harm in including Flex component by default and making just the `<ButtonIcon/>` conditional? Then we could skip the ternary and the `<Flex/>` seems like it could come it handy.

src/Button/Button.js
```diff
@@ -71,34 +81,104 @@ const StyledButton = createComponent({
...
+          <ButtonIcon name={props.icon} sizing={props.size} hasText={hasText} />
```
cehsu: How about passing the ButtonIcon component as an `icon` prop to Button? All of its props would still be accessible via: `props.icon.props.xyz`, and then Button doesn't need to worry about as many of the icon details and users can more easily customize spacing, for example, if they need to?

src/theme.js
mathewmorris: Well, I think we could probably come up with a better variable name. The purpose of `textColor` is an override for when the button is a Text Button. For the case of the `primary` button, `fontColor` is white. that won't work for a Text Button. 

I thought about using a function to determine this, but in case someone actually wants a dark background and a white Text Button color, it would be best to leave it as a overridable variable.

src/Button/Button.js
mathewmorris: Yes! :) I liked this from the Input PR. I will definitely make those changes, because it's WAY better than being so verbose about all this.

src/Button/Button.js
```diff
@@ -71,34 +80,96 @@ const StyledButton = createComponent({
...
-        border-color: ${lighten(0.05, backgroundColor)};
+        background-color: ${backgroundColor};
+        border-color: ${borderColor || backgroundColor};
+        color: ${fontColor};
+
+        ${hover &&
+          css`
+            background-color: ${hover.backgroundColor || backgroundColor};
+            border-color: ${hover.borderColor || hover.backgroundColor || backgroundColor};
+            color: ${hover.fontColor || fontColor};
```
erikshestopal: When hovering over a text only button, I'm not seeing the text color change per the Figma specs. Same things goes for the `active` state.

src/theme.js
```diff
@@ -151,15 +231,26 @@ export default (overrides = {}) => {
     },
   };
 
+  const fonts = [
+    {
+      name: 'Tiempos',
+      url: '//cdn.heydoctor.com/fonts/TiemposHeadlineWeb-Semibold.woff2',
```
kylealwyn: I'm not sure if we need to include this in the theme - could you elaborate on how you seeing this used in the consuming apps?

src/theme.js
kylealwyn: I agree with @cehsu here - fontColor and textColor will be too easy to conflate. I propose text buttons be their own variant, e.g. `greyText`, so we don't have to be too tricky with the `text` prop

src/Badge/__snapshots__/Badge.spec.js.snap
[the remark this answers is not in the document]
kylealwyn: Let's figure out how to normalize these values

src/Button/Button.js
```diff
@@ -71,34 +80,96 @@ const StyledButton = createComponent({
...
+        `}
+
       ${space};
     `;
   },
 });
 
-const Button = React.forwardRef((props, ref) => <StyledButton {...props} ref={ref} />);
+const Button = React.forwardRef((props, ref) => {
+  const hasText = !!props.children;
```
kylealwyn: No need to set this extra variable - can look at `props.children` directly in `StyledButton`

src/Button/Button.js
```diff
@@ -71,34 +80,96 @@ const StyledButton = createComponent({
...
       width: ${block ? '100%' : 'auto'};
-      background: ${outline || transparent ? 'transparent' : backgroundColor};
-      border: ${transparent ? 'none' : `1px solid ${backgroundColor}`};
+      background: ${outline ? 'transparent' : backgroundColor};
+      border-color: ${outline ? backgroundColor : borderColor || backgroundColor};
+      border-style: solid;
+      border-width: 1px;
       transition: 175ms;
+      white-space: nowrap;
+      outline: none;
```
kylealwyn: I don't think we want this for accessibility reasons

src/Button/Button.js
```diff
@@ -107,7 +178,8 @@ Button.propTypes = {
   block: PropTypes.bool,
   disabled: PropTypes.bool,
   loading: PropTypes.bool,
-  transparent: PropTypes.bool,
+  icon: PropTypes.element,
```
kylealwyn: I think it may be more ergonomic if this was a string so we don't have to import the Icon component every time we want to use: `<Button leftIcon="alert" leftIconProps={{ size: 48 }} />`

src/theme.js
mathewmorris: The goal here was to default as much as possible to HeyDoctor/GoodRx design, but I can understand leaving this out for the heydoctor theme file in each project.

jlemm45: @kylealwyn It was just added to provide it as a default font. If we don't want to add it to the library we can add it individually to each app.

src/Button/Button.js
jlemm45: @kylealwyn I updated to support this pattern. Going by your example I went ahead and added support for `leftIcon` & `rightIcon`.

erikshestopal:
Echoing what Claire brough up [here](https://github.com/heydoctor/molekule/pull/66#discussion_r316375738), are we still going to support these types of alerts? 

Screen Shot 2019-08-21 at 3 21 48 PM

kylealwyn:
No

PR #52, opened 2019-08-02
feat(Modal): Accessibility updates, width constraint

mathewmorris:
This update is for accessibility and width constraints as noted in design specs.

Replaces `react-focus-lock` library with `react-focus-on` which gives the added accessibility features

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -1,14 +1,13 @@
-import React, { createContext, useContext, useEffect, useState } from 'react';
+import React, { createContext, useContext, useEffect, useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { keyframes, css } from 'styled-components';
 import * as animations from 'react-animations';
 import { Transition } from 'react-transition-group';
-import FocusLock from 'react-focus-lock';
+import { FocusOn } from 'react-focus-on';
 import Portal from '../Portal';
 import Flex from '../Flex';
 import Box from '../Box';
 import Icon from '../Icon';
-import { useKeyPress } from '../hooks';
 import { createComponent, themeGet } from '../utils';
 
 const ModalContext = createContext({});
@@ -77,6 +76,7 @@ const ModalContent = createComponent({
 
 function Modal({ children, title, animationDuration, showClose, onClose, open, ...props }) {
   const [isOpen, setOpen] = useState(open);
+  const modal = useRef();
 
   const handleClose = () => {
     setOpen(false);
@@ -84,17 +84,6 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
   };
 
   const handleContentClick = event => event.stopPropagation();
-
-  const handleBackdropClick = () => {
-    if (!props.closeOnBackdropClick) return;
-
-    handleClose();
-  };
-
-  useKeyPress('Escape', () => {
-    if (!isOpen || !props.closeOnEscape) return;
-    handleClose();
-  });
 
   useEffect(() => {
     if (open !== isOpen) {
@@ -107,13 +96,18 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
       <Portal>
         <Transition in={isOpen} timeout={animationDuration}>
           {state => (
-            <Backdrop transitionState={state} onClick={handleBackdropClick}>
-              <FocusLock lockProps={{ style: { maxHeight: '100%' } }} disabled={!isOpen}>
-                <ModalContent transitionState={state} onClick={handleContentClick} {...props}>
+            <Backdrop transitionState={state}>
+              <FocusOn
+                onClickOutside={handleClose}
+                onEscapeKey={handleClose}
+                lockProps={{ style: { maxHeight: '100%' } }}
+                enabled={isOpen}
+                shards={[modal]}>
+                <ModalContent transitionState={state} onClick={handleContentClick} ref={modal} {...props}>
                   {title && <Modal.Header title={title} showClose={showClose} />}
                   {children}
                 </ModalContent>
-              </FocusLock>
+              </FocusOn>
             </Backdrop>
           )}
         </Transition>
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -82,6 +82,7 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
   const handleClose = () => {
     setOpen(false);
     onClose();
+    toggleBlur();
   };
 
   const handleContentClick = event => event.stopPropagation();
@@ -93,7 +94,9 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
   };
 
   const toggleBlur = () => {
-    blurNode.style.filter = open ? 'blur(1px)' : null;
+    if (blur && blurNode) {
+      blurNode.style.filter = open ? 'blur(1px)' : null;
+    }
   };
 
   useKeyPress('Escape', () => {
@@ -106,14 +109,12 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
       setOpen(open);
     }
 
-    if (blur && blurNode) {
-      toggleBlur();
-    }
+    toggleBlur();
   }, [open]);
 
   useEffect(() => {
     if (blur) {
-      const target = document.getElementById(blurTarget);
+      const target = document.querySelector(blurTarget);
 
       if (target) {
         setBlurNode(target);
@@ -127,7 +128,7 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
         };
       }
     }
-  }, []);
+  }, [blur, blurTarget]);
 
   return (
     <ModalContext.Provider value={{ handleClose }}>
@@ -176,7 +177,7 @@ Modal.defaultProps = {
   animationDuration: 175,
   onClose: () => {},
   blur: false,
-  blurTarget: 'root',
+  blurTarget: '#root',
 };
 
 Modal.Title = createComponent({
@@ -238,8 +239,7 @@ Modal.Body = createComponent({
 
 Modal.Footer = createComponent({
   name: 'ModalFooter',
-  style: ({ theme }) => css`
-    background: ${theme.colors.grayLightest};
+  style: css`
     border-bottom-left-radius: ${themeGet('radius')}px;
     border-bottom-right-radius: ${themeGet('radius')}px;
   `,
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -1,4 +1,4 @@
-import React, { createContext, useContext, useEffect, useState } from 'react';
+import React, { createContext, useContext, useEffect, useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { keyframes, css } from 'styled-components';
 import * as animations from 'react-animations';
@@ -77,7 +77,7 @@ const ModalContent = createComponent({
 
 function Modal({ children, title, animationDuration, showClose, onClose, open, blur, blurTarget, ...props }) {
   const [isOpen, setOpen] = useState(open);
-  const [blurNode, setBlurNode] = useState();
+  const blurNode = useRef();
 
   const handleClose = () => {
     setOpen(false);
@@ -94,8 +94,8 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
   };
 
   const toggleBlur = () => {
-    if (blur && blurNode) {
-      blurNode.style.filter = open ? 'blur(1px)' : null;
+    if (blur && blurNode.current) {
+      blurNode.current.style.filter = open ? 'blur(1px)' : null;
     }
   };
 
@@ -117,14 +117,16 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, b
       const target = document.querySelector(blurTarget);
 
       if (target) {
-        setBlurNode(target);
-        target.style.transitionDuration = '750ms';
-        target.style.overflow = 'auto';
+        blurNode.current = target;
+
+        blurNode.current.style.transitionDuration = '750ms';
+        blurNode.current.style.overflow = 'auto';
+        blurNode.current.style.zIndex = 0;
 
         return () => {
-          target.style.transitionDuration = null;
-          target.style.filter = null;
-          target.style.overflow = null;
+          blurNode.current.style.transitionDuration = null;
+          blurNode.current.style.filter = null;
+          blurNode.current.style.overflow = null;
         };
       }
     }
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -175,7 +175,7 @@ Modal.Header = ({ title, children, showClose = true }) => {
 
           {showClose && (
             <Box ml="auto">
-              <Icon name="close" color="greyDark" style={{ cursor: 'pointer' }} size={16} onClick={handleClose} />
+              <Icon name="close" color="greyDark" style={{ cursor: 'pointer' }} size={24} onClick={handleClose} />
             </Box>
           )}
         </Flex>
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -24,8 +24,8 @@ const Backdrop = createComponent({
     z-index: 1000;
     padding: 1rem;
     display: flex;
-    align-items: center;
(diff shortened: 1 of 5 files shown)


erikshestopal:
@mathewmorris can you please elaborate (provide more context) on why these changes are necessary and what issues they're solving? For example, why do we need a `blurTarget`? thanks! 😄 👍

src/Modal/Modal.js
```diff
@@ -197,14 +232,13 @@ Modal.Header = ({ title, children, showClose = true }) => {
...
-    padding: 1.25rem;
+    padding-bottom: 24px;
   `,
 });
 
 Modal.Footer = createComponent({
   name: 'ModalFooter',
   style: ({ theme }) => css`
-    padding: 1rem 1.25rem;
     background: ${theme.colors.grayLightest};
```
erikshestopal: Curious why we have this `background` here...Was this intentional? 

Screen Shot 2019-08-06 at 12 50 36 PM

src/Modal/Modal.js
```diff
@@ -98,8 +105,30 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
+        target.style.overflow = 'auto';
+
+        return () => {
+          target.style.transitionDuration = null;
+          target.style.filter = null;
+          target.style.overflow = null;
+        };
+      }
+    }
+  }, []);
```
erikshestopal: Do we need to add `[blur, blurTarget]` to the deps array since we want the effect to synchronize if those props change?

src/Modal/Modal.js
```diff
@@ -98,8 +105,30 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
     }
+
+    if (blur && blurNode) {
+      toggleBlur();
+    }
   }, [open]);
 
+  useEffect(() => {
+    if (blur) {
+      const target = document.getElementById(blurTarget);
```
erikshestopal: Maybe we should use `document.querySelector` to provide a more flexible API? e.g someone might want to use a class name, etc

mathewmorris:
@erikshestopal I added a little blurb to the description of this PR :)

src/Modal/Modal.js
```diff
@@ -73,8 +75,9 @@ const ModalContent = createComponent({
   `,
 });
 
-function Modal({ children, title, animationDuration, showClose, onClose, open, ...props }) {
+function Modal({ children, title, animationDuration, showClose, onClose, open, blur, blurTarget, ...props }) {
   const [isOpen, setOpen] = useState(open);
+  const [blurNode, setBlurNode] = useState();
```
erikshestopal: Have you considered using a ref to store the `blurTarget` since it will be a DOM node? I don't think we need to re-render the component using `setBlurNode` because all we want to do is keep track of this node across renders, which `useRef` is designed to do. LMK your thoughts.

src/Modal/Modal.js
mathewmorris: We want to mimic a `componentDidMount` lifecycle method here. Which is why there is an empty deps array. If we add `[blur, blurTarget]` as the deps array, then it will keep grabbing applying those styles to the `#root` on every render. 

Those props shouldn't be changed frequently.

src/Modal/Modal.js
mathewmorris: True. That would be more flexible.

src/Modal/Modal.js
erikshestopal: >  If we add [blur, blurTarget] as the deps array, then it will keep grabbing applying those styles to the #root on every render.

Correct me if I'm wrong, but `#useEffect` will only reapply those styles ***if*** the deps array changes. If `blur` and `blurTarget` do not change between renders, `useEffect` will not re-run.

src/Modal/Modal.js
mathewmorris: What do you mean exactly? I've only used refs for nodes created by the component, is there a way to use `createRef` to target a node outside of the component?

src/Modal/Modal.js
mathewmorris: You're 100% correct. Did some testing to verify and the cleanup still fires too. Changing to your suggestion :) :)

src/Modal/Modal.js
mathewmorris: Ha. Removing.

src/Modal/Modal.js
```diff
@@ -73,12 +75,14 @@ const ModalContent = createComponent({
   `,
 });
 
-function Modal({ children, title, animationDuration, showClose, onClose, open, ...props }) {
+function Modal({ children, title, animationDuration, showClose, onClose, open, blur, blurTarget, ...props }) {
```
erikshestopal: Let's default `blur = true` so that blur is opt-out rather than having to specify `blur` on every single modal in the web-app.

mathewmorris: 👍

src/Modal/Modal.js
erikshestopal:
```jsx
function DumbModal() {
  const blurNode = useRef();

  const toggleBlur = () => {
    if (blur && blurNode.current) {
      blurNode.current.style.filter = open ? 'blur(1px)' : null;
    }
  };

  useEffect(() => {
    const target = document.querySelector(blurTarget);

    if (target) {
      blurNode.current = target;
    }
  }, []);

}
```

erikshestopal:
@mathewmorris ok sounds good 👍

src/Modal/Modal.js
mathewmorris: Hmm. Makes sense. Refs should be used if needing to touch the DOM too. Great suggestion!

erikshestopal:
The animations in Firefox seem a little slow right now compared with the existing ones. Maybe this has something to do with overriding `style.transitionDuration`? 

**Before**
prev-slow-ffox

**After**
slow-ffox

src/Modal/Modal.js
```diff
@@ -98,8 +108,27 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
+      const target = document.querySelector(blurTarget);
+
+      if (target) {
+        blurNode.current = target;
+
+        blurNode.current.style.overflow = 'auto';
+
+        return () => {
+          blurNode.current.style.filter = null;
+          blurNode.current.style.overflow = null;
```
cehsu: How about keeping track of the initial values for `overflow` and `filter`, so that they can be restored here and when the filter is being set?

src/Modal/Modal.js
```diff
@@ -73,12 +75,14 @@ const ModalContent = createComponent({
   `,
 });
 
-function Modal({ children, title, animationDuration, showClose, onClose, open, ...props }) {
+function Modal({ children, title, animationDuration, showClose, onClose, open, blur, blurTarget, ...props }) {
   const [isOpen, setOpen] = useState(open);
+  const blurNode = useRef();
```
cehsu: We could initialize the ref value here:
```suggestion
  const blurNode = useRef(document.querySelector(blurTarget));
```

src/Modal/Modal.js
```diff
@@ -89,6 +93,12 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
     handleClose();
   };
 
+  const toggleBlur = () => {
+    if (blur && blurNode.current) {
+      blurNode.current.style.filter = open ? 'blur(1px)' : null;
```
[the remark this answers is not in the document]
mathewmorris: I tweaked a little. Works great. Makes sense. 👍

src/Modal/Modal.js
```diff
@@ -100,6 +103,32 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
   }, [open]);
 
+  useEffect(() => {
+    if (blur && blurNode.current) {
+      const isBackgroundBlurred = !!blurNode.current.style.filter;
+      if (isOpen !== isBackgroundBlurred) {
+        blurNode.current.style.filter = isOpen ? 'blur(1px)' : null;
+      }
+    }
+  }, [isOpen]);
```
erikshestopal: we'll want to add `blur` to deps because the prop _could_ change between renders. 
```suggestion
  }, [isOpen, blur]);
```

mathewmorris:
@cehsu Well, now I'm realizing this filter isn't working out very well. When adding `filter: blur()` onto the root element, it's causing all children to become relative to it. So, all fixed items are no longer fixed to the viewport (ie: the side menu). I'm not sure of a way to make the entire background blur without causing this to happen. The why is explained in 
[Stacking Context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context).

I'm not sure we would want to introduce something to Molekule that could have unexpected results due to this. Do you have any ideas other than using `filter: blur()` to achieve the same effect?

blur-modal-stack-context

erikshestopal:
@mathewmorris Can we use [backdrop-filter](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter) applied to the modal backdrop? No FF support but seems to be supported by Chrome, Safari and Edge.

mathewmorris:
@erikshestopal I did see that, but because of the FF not supporting, I nixed it. And looks like FF nightly team is working on support for it. Latest movement was a few hours ago. If you guys are okay with it, then I can make it happen. It would fix that context stack issue as well.

src/Modal/Modal.js
```diff
@@ -138,8 +140,8 @@ Modal.defaultProps = {
   showClose: true,
   closeOnBackdropClick: true,
   closeOnEscape: true,
-  minWidth: 250,
-  maxWidth: 768,
+  minWidth: 350,
+  maxWidth: 350,
```
erikshestopal: What prompted this `maxWidth` change to `350`? The camera modal in the `web-app` takes up a lot more space than `350` and I think we want to keep it that way. 😄

mathewmorris: The camera modal is the only exception in this case. Which is why I decided to change the default `maxWidth` to 350 which is what Jay specified [here](https://app.asana.com/0/1129589982637900/1132970972936105).

All modals can override this in their props.

cehsu:
It seems like we could retain the blur, and leave the scroll lock for later.

cehsu:
Depending on how disruptive the lack of FF support is, it’s worth checking out.

mathewmorris:
GoodRx actually wants IE support (Hurshal filled us in).

The real problem with the `filter: blur` attribute, it causes problems with the stacking context, specifically and absolutely or fixed position elements.

And then the `backdrop-blur` will probably never be supported in IE. 

But FF will have support for `backdrop-blur` later this year, and IE never supporting is fine, I would say `backdrop-blur` is the way to go. Otherwise, leave it out.

cehsu:
I did a quick test and managed to achieve the desired blur and scroll lock behavior by specifying multiple blur nodes, and using [`react-focus-on`](https://github.com/theKashey/react-focus-on) (as recommended by the [focus lock library](https://github.com/theKashey/react-focus-lock#more) that we use), respectively.

We could leave the blur node specifics up to users, e.g.
```
const blurNodes = [
'document.querySelector('#root').firstChild.firstChild.firstChild',
'document.querySelector('#root').firstChild.childNodes[2]',
];
...
I'll double check cross browser compatibility.
blurNodes={blurNodes}
...
```

src/Modal/Modal.js
```diff
@@ -60,6 +59,8 @@ const ModalContent = createComponent({
     background-clip: padding-box;
     box-shadow: 0 8px 30px rgba(0, 29, 54, 0.1);
     border-radius: ${themeGet('radius')}px;
+    padding: 16px;
+    font-size: 14px;
```
kylealwyn: Why does this need to be set explicitly?

src/Modal/Modal.js
```diff
@@ -105,13 +96,18 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
       <Portal>
         <Transition in={isOpen} timeout={animationDuration}>
           {state => (
-            <Backdrop transitionState={state} onClick={handleBackdropClick}>
-              <FocusLock lockProps={{ style: { maxHeight: '100%' } }} disabled={!isOpen}>
-                <ModalContent transitionState={state} onClick={handleContentClick} {...props}>
+            <Backdrop transitionState={state}>
+              <FocusOn
```
kylealwyn: What prompted the new library? Can we remove the `react-focus-lock` library so we don't have redundant libs doing similar things?

mathewmorris:
@cehsu Awesome! Glad that's working well on your end. Same here.

I am currently working on a `blurNodes` solution. It is working, but I don't have a way to reliably target the Docz `menu` and `main` divs since they are using some styling library that will randomize the class name :(

src/Modal/Modal.js
```diff
@@ -150,16 +146,18 @@ Modal.Title = createComponent({
...
   `,
 });
 
 const ModalHeader = createComponent({
   name: 'ModalHeader',
   style: css`
-    font-size: 1.5rem;
-    padding: 1rem 1.25rem 0;
+    font-size: 16px;
+    font-weight: 700;
```
kylealwyn: Why do we need to add font properties to this containing element?

src/Modal/Modal.js
```diff
@@ -105,13 +96,18 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
-              <FocusLock lockProps={{ style: { maxHeight: '100%' } }} disabled={!isOpen}>
-                <ModalContent transitionState={state} onClick={handleContentClick} {...props}>
+            <Backdrop transitionState={state}>
+              <FocusOn
+                onClickOutside={handleClose}
+                onEscapeKey={handleClose}
+                lockProps={{ style: { maxHeight: '100%' } }}
+                enabled={isOpen}
+                shards={[modal]}>
+                <ModalContent transitionState={state} onClick={handleContentClick} ref={modal} {...props}>
```
kylealwyn: Do we need the `handleContentClick` functionality anymore?

src/Modal/Modal.js
```diff
@@ -83,35 +86,61 @@ function Modal({ children, title, animationDuration, showClose, onClose, open, .
...
   useEffect(() => {
     if (open !== isOpen) {
       setOpen(open);
     }
   }, [open]);
 
+
+  useEffect(() => {
+    if (blur) {
+      const targets = document.querySelectorAll(blurTargets);
```
kylealwyn: Let's drop the blur - this is overly complex for a small luxury with less than great performance and support

src/Modal/Modal.js
mathewmorris: 👍 I believe the main goal was accessibility. `react-focus-on` creates an accessible experience. So, now the modal and dropdown should be more accessible. I went ahead and removed the `react-focus-lock` library, and am now using `react-focus-lock` exclusively. (Which is maintained by the same group, and also includes `react-focus-lock` in it's dependencies.)

src/Modal/Modal.js
```diff
@@ -185,7 +175,7 @@ Modal.Header = ({ title, children, showClose = true }) => {
 
           {showClose && (
             <Box ml="auto">
-              <Icon name="close" color="greyDark" style={{ cursor: 'pointer' }} onClick={handleClose} />
+              <Icon name="close" color="greyDark" style={{ cursor: 'pointer' }} size={16} onClick={handleClose} />
```
kylealwyn: I think this may need to be bumped up to 24, looking a little small 

Screen Shot 2019-08-21 at 2 38 39 PM

cehsu:
@mathewmorris could we use the selector in the example?
```
'document.querySelector('#root').firstChild.firstChild.firstChild',
'document.querySelector('#root').firstChild.childNodes[2]',
```

kylealwyn:
@cehsu What example are you referring to? I'd prefer we refrain from using that anywhere - it's too brittle and specific. The blur functionality was removed due to the unnecessary complexity it was adding.

cehsu:
@kylealwyn see https://github.com/heydoctor/molekule/pull/52#issuecomment-522748810.

cehsu:
If we want to allow users to add a blur filter, but don't want to manage it, we could add an `onOpen` prop, instead, so users can add whatever they need.

erikshestopal:
@mathewmorris can you please cherrypick 82ed906 to fix the netlify build? Kind weird that these builds started failing all of sudden 🤔

PR #56, opened 2019-08-10
Async Formbot Validations

erikshestopal:
## Why? 
Our validation schema library `yup` allows for sync as well as async validations and we've not been able to utilize this feature due to the constraint of `Formbot` validating forms synchronously. Most recently, Scott on the `web-app` and myself on the `mobile-app` have run into an issue where we needed to hit an API endpoint as part of validating a patient's address (zip + state matching) but making this work required some ninjitsu (utilizing local state, effects to run the validations, etc). With this proposal, all of that goes away and we can define smarter validation schemas that support async validations. 

## Architecture 
There are no breaking changes in the PR so we can adopt as needed, but we now allow the ability to pass a function to `validationSchema`. The function will be called with one argument, the Formbot context and expects a return object with a `yup` schema definition. 

```jsx
const createSchema = ({ values: { state } }) => ({
  state: yup
    .string()
    .required()
    .test('valid-state', 'Not available in your state!', async value => {
      const isValidState = await validateServiceAvailability({ state: value });
      return isValidState;
    }),
  zip: yup
    .string()
    .min(5)
    .required()
    .test('valid-zip', 'Zip does not match state', async zip => {
      const stateFromZip = await getStateFromZip(zip);
      return stateFromZip === state;
    }),
});

<Formbot validationSchema={createSchema}>
  <Input name="state" />
  <Input name="zip" />
</Formbot>;
```

## What Else?
I've also refactored `FormError` to be a be a much smarter component. Sometimes we want to render an error that is outside of an input (something that spans an entire row, for example, instead of just showing up underneath a field). `FormError` is now connected to the `Formbot` context and can render the appropriate error given a `name` prop that corresponds to the error key. 

```jsx
<Formbot validationSchema={createSchema}>
  <Input name="state" renderError={false} />
  <FormError name="state" /> // default error
  <FormError name="state">{error => <span>My custom error {error}</span>}</FormError>;
</Formbot>;
```

### DIFF
--- a/src/Form/EasyInput.js
+++ b/src/Form/EasyInput.js
@@ -8,7 +8,7 @@ import { Context } from './Formbot';
  */
 const PureInput = React.memo(({ Component, ...props }) => <Component {...props} />);
 
-function EasyInput({ name, Component, ...props }) {
+function EasyInput({ name, Component, renderError = true, ...props }) {
   const state = useContext(Context);
 
   if (!state) {
@@ -25,7 +25,7 @@ function EasyInput({ name, Component, ...props }) {
     <PureInput
       name={name}
       value={value !== undefined ? value : defaultValue}
-      error={state.errors[name]}
+      error={renderError ? state.errors[name] : undefined}
       onChange={state.onChange}
       onBlur={state.onBlur}
       onFocus={state.onFocus}
@@ -36,6 +36,6 @@ function EasyInput({ name, Component, ...props }) {
 }
 
 export const createEasyInput = Component =>
-  forwardRef((props, ref) => <EasyInput Component={Component} forwardedRef={ref} {...props} />);
+  forwardRef((props, ref) => <EasyInput Component={Component} forwardedRef={ref} {...props} />); // eslint-disable-line
 
 export default EasyInput;
--- a/src/Form/EasyInput.js
+++ b/src/Form/EasyInput.js
@@ -8,7 +8,7 @@ import { Context } from './Formbot';
  */
 const PureInput = React.memo(({ Component, ...props }) => <Component {...props} />);
 
-function EasyInput({ name, Component, renderError = true, ...props }) {
+function EasyInput({ name, Component, shouldRenderError = true, ...props }) {
   const state = useContext(Context);
 
   if (!state) {
@@ -25,7 +25,7 @@ function EasyInput({ name, Component, renderError = true, ...props }) {
     <PureInput
       name={name}
       value={value !== undefined ? value : defaultValue}
-      error={renderError ? state.errors[name] : undefined}
+      error={shouldRenderError ? state.errors[name] : undefined}
       onChange={state.onChange}
       onBlur={state.onBlur}
       onFocus={state.onFocus}
--- a/src/Form/Formbot.js
+++ b/src/Form/Formbot.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { omit } from 'lodash';
 
 export const Context = React.createContext(null);
 
@@ -26,7 +27,7 @@ export default class Formbot extends React.Component {
   static propTypes = {
     initialValues: PropTypes.shape(),
     validations: PropTypes.shape(),
-    validationSchema: PropTypes.oneOfType([PropTypes.shape(), PropTypes.func]),
+    validationSchema: PropTypes.shape(),
     onFocus: PropTypes.func,
     onChange: PropTypes.func,
     onBlur: PropTypes.func,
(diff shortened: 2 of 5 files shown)


src/Form/Formbot.js
```diff
@@ -111,11 +114,17 @@ export default class Formbot extends React.Component {
     });
   };
 
-  makeValidation = (validation, value) => {
-    if (!validation.validator(value)) {
-      throw new Error(validation.message || 'This field is required');
-    }
-  };
+  setStateErrors = (field, error, cb) =>
```
kylealwyn: Could we change to `setErrors` to be consistent with `setValues`?

src/Form/Formbot.js
```diff
@@ -125,8 +134,11 @@ export default class Formbot extends React.Component {
         return;
       }
 
-      const fromSchema = !!this.props.validationSchema;
-      const validation = (fromSchema ? this.props.validationSchema : this.props.validations)[field];
+      const { validationSchema, validations } = this.props;
+
+      const fromSchema = !!validationSchema;
+      const schema = typeof validationSchema === 'function' ? validationSchema(this.getContext()) : validationSchema;
```
kylealwyn: This looks like it may be a slight performance drain to regenerate schema for every field validation. Can we look at providing the Formbot context as the context field in the yup options object we can pass to the validate functions? https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror

src/Form/Formbot.js
```diff
@@ -138,7 +150,15 @@ export default class Formbot extends React.Component {
 
       try {
         if (fromSchema) {
-          validation.validateSync(fieldValue, { values: this.state.values });
+          if (typeof validationSchema === 'function') {
```
kylealwyn: I think we'll want to check if `validation` is a function here

src/Form/Formbot.js
```diff
@@ -111,11 +114,17 @@ export default class Formbot extends React.Component {
...
 
-  makeValidation = (validation, value) => {
-    if (!validation.validator(value)) {
-      throw new Error(validation.message || 'This field is required');
-    }
-  };
+  setStateErrors = (field, error, cb) =>
+    this.setState(
+      state => ({
+        ...state,
```
kylealwyn: I don't think we need to spread existing state here

src/Form/EasyInput.js
```diff
@@ -25,7 +25,7 @@ function EasyInput({ name, Component, ...props }) {
     <PureInput
       name={name}
       value={value !== undefined ? value : defaultValue}
-      error={state.errors[name]}
+      error={renderError ? state.errors[name] : undefined}
```
kylealwyn: Seems like we've generally established a pattern that `render*` should be a render function. Maybe prefix with should?

src/Form/Formbot.js
```diff
@@ -125,8 +134,11 @@ export default class Formbot extends React.Component {
         return;
       }
 
-      const fromSchema = !!this.props.validationSchema;
-      const validation = (fromSchema ? this.props.validationSchema : this.props.validations)[field];
+      const { validationSchema, validations } = this.props;
+
+      const fromSchema = !!validationSchema;
```
kylealwyn: semantic nit: `hasSchema`

src/Form/Formbot.js
```diff
@@ -137,8 +149,16 @@ export default class Formbot extends React.Component {
...
+        if (hasSchema) {
+          if (typeof validation.validate === 'function' && isAsyncValidation) {
+            validation.validate(fieldValue, validationOpts).catch(e => {
+              this.setErrors({ [field]: e.message }, resolve);
+            });
+
+            return;
+          }
+
+          validation.validateSync(fieldValue, validationOpts);
```
schlegz: Is it necessary to make a distinction between sync/async validations?  I'm assuming we could always call `validation.validate` with a sync or async function and have it resolve correctly.  I did a quick test in the `molekule` example server setting `async: true` in the schema and passing a non-async function to `test()`.  It still validates correctly so that is why I'm thinking that maybe this logic could be simplified.  Let me know if i'm understanding this incorrectly.

erikshestopal: Good observation, will fix 👍

src/Form/Formbot.js
```diff
@@ -137,8 +148,12 @@ export default class Formbot extends React.Component {
       let errorMsg;
 
       try {
-        if (fromSchema) {
-          validation.validateSync(fieldValue, { values: this.state.values });
+        if (hasSchema && typeof validation.validate === 'function') {
+          validation.validate(fieldValue, validationOpts).catch(e => {
+            this.setErrors({ [field]: e.message }, resolve);
```
schlegz: Why is it necessary to catch the error here and not let it be caught and handled in the catch/finally blocks below?

erikshestopal: @schlegz It's a promise and the error is generated not when the `try/catch` block is running but later so there's no way to catch that promise rejection with a `try/catch`. 


```js
const error = async () => throw new Error();

const validate = () => new Promise((resolve) => {
  try {
    error();
    resolve();
  } catch (e) { 
    //  promise rejection will not be caught here
  }
})
```

PR #57, opened 2019-08-13
fix(validations): Fix handling of async validations as well as order of onChange notification

schlegz:
It was discovered when refactoring the `AddressForm` component that the `onChange` notification in `FormBot` was called before field updating/validation was complete.  This is an issue if in an `onChange` handler for `FormBot` you are expecting to test for form validity (i.e. updating your UI according to the validity of a recent change to form data).  Additionally the `return` from the async validation block was triggering the `finally` handler causing premature promise resolution and triggering the `onChange` handler before validations were complete.

### DIFF
--- a/src/Form/Formbot.js
+++ b/src/Form/Formbot.js
@@ -210,9 +210,7 @@ export default class Formbot extends React.Component {
       () => {
         this.updateField(field, { validated: false })
           .then(() => this.validateField(field))
-          .then(() => {
-            this.props.onChange(field, value, this.state.values);
-          })
+          .then(() => this.props.onChange(field, value, this.state.values))
       }
     );
   };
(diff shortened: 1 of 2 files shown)


src/Form/Formbot.js
```diff
@@ -200,10 +208,11 @@ export default class Formbot extends React.Component {
...
       },
       () => {
-        this.props.onChange(field, value, this.state.values);
-        this.updateField(field, { validated: false }).then(() => {
-          this.validateField(field);
-        });
+        this.updateField(field, { validated: false })
+          .then(() => this.validateField(field))
+          .then(() => {
+            this.props.onChange(field, value, this.state.values);
```
erikshestopal: super nit: add implicit return to be consistent with L212

schlegz:
@choochootrain Are you talking about what is generated by `docz` or in `Formbot.example.js`?  How would I update what is generated by `docz`?

PR #58, opened 2019-08-14
fix: Use updater function in setState for Formbot updates

erikshestopal:
## Summary 
Using a function as a state updater seems to fix an issue with stale state data preventing a validation function from running.

### DIFF
--- a/src/Form/Formbot.js
+++ b/src/Form/Formbot.js
@@ -74,17 +74,16 @@ export default class Formbot extends React.Component {
     }));
   }
 
-  setValues(values = {}) {
+  setValues = (values = {}) =>
     this.setState(
-      {
+      state => ({
         values: {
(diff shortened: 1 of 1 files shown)


src/Form/Formbot.js
```diff
@@ -97,24 +96,21 @@ export default class Formbot extends React.Component {
...
+    new Promise(resolve => {
       this.setState(
-        {
+        state => ({
           fields: {
-            ...this.state.fields,
+            ...state.fields,
             [field]: {
-              ...fieldState,
+              ...state.fields[field],
```
cehsu: Does this no longer require the default `{}` value?

erikshestopal: @cehsu 
```
{ ...undefined, ...null } == {}
```

PR #61, opened 2019-08-16
fix(FormBot): Fix assumption that validate is async

schlegz:
Fix for https://sentry.io/organizations/heydoctor/issues/1164080252.

Should not assume `validate` is chainable.

### DIFF
--- a/src/Form/Formbot.js
+++ b/src/Form/Formbot.js
@@ -146,16 +146,15 @@ export default class Formbot extends React.Component {
       let errorMsg;
 
       if (hasSchema && typeof validation.validate === 'function') {
-        validation
-          .validate(fieldValue, validationOpts)
-          .catch(e => {
-            errorMsg = e.message;
-          })
-          .finally(() => {
-            this.updateField(field, { validated: true }).then(() => {
-              this.setErrors({ [field]: errorMsg }, resolve);
-            });
+        try {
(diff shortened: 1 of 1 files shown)


src/Form/Formbot.js
```diff
@@ -146,16 +146,15 @@ export default class Formbot extends React.Component {
...
-          .catch(e => {
-            errorMsg = e.message;
-          })
-          .finally(() => {
-            this.updateField(field, { validated: true }).then(() => {
-              this.setErrors({ [field]: errorMsg }, resolve);
-            });
+        try {
+          validation.validate(fieldValue, validationOpts);
+        } catch(e) {
```
erikshestopal: I don't think this is going to work actually `async` functions. This will be rejected with unhandled promise error because the `try/catch` block won't catch the promise rejection. We need to check if the result of the calling the function returns a promise or whether is just returns a value or throws.

schlegz: @erikshestopal Turns out this is a non-issue, the user was on an unsupported version of the browser.

choochootrain:
https://sentry.io/organizations/heydoctor/issues/1164080252/?referrer=slack

this came from chrome 49 which is super old so not worth fixing. task to lock out older browsers here:
https://app.asana.com/0/1127722639002160/1135988081676913/f

PR #62, opened 2019-08-20
fix: Remove support for multiple input sizes

kylealwyn:
Confirmed with Jay. We only use one size for Inputs. Any other requirements for special sizing can be handled on a case by case basis.

### DIFF
--- a/src/Form/Select.js
+++ b/src/Form/Select.js
@@ -7,23 +7,23 @@ import Icon from '../Icon';
 import Flex from '../Flex';
 import Label from './Label';
 import { createEasyInput } from './EasyInput';
-import { createComponent } from '../utils';
+import { themeGet, createComponent } from '../utils';
 
 const SelectContainer = createComponent({
   name: 'SelectContainer',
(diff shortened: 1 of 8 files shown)


src/Form/Select.js
```diff
@@ -7,23 +7,23 @@ import Icon from '../Icon';
...
 
 const SelectContainer = createComponent({
   name: 'SelectContainer',
   as: Flex,
-  style: ({ size, theme, value, borderRadius = theme.radius }) => css`
+  style: ({ theme, value }) => css`
     background: white;
     border: 1px solid ${theme.colors.greyLight};
-    height: ${theme.heights[size]}px;
+    height: 48px;
```
cehsu: Do we want to reuse variables to hardcode this size via `themeGet('heights.xl')px;` here and for other size magic string above (e.g. font-size), even though they are no longer configurable?

kylealwyn: No - I want to move away from heights in totality. It's too brittle.

PR #66, opened 2019-08-21
feat: Button variant styling proposal

kylealwyn:
## Summary

The logic to determine Button styles felt slightly cumbersome. Rather than trying to be too tricky on how to handle different button styles, e.g. outlines, text only, colors, I propose we configure every variation as a variant, with naming conventions such as `colorText` or `colorOutline`. The `variant` blocks in the theme are now fully compatible `styled-components` objects. 

They can also be defined as a function which can receive the button props, in case any extra logic is needed to determine styling.

Screen Shot 2019-08-21 at 12 47 30 PM

### DIFF
--- a/src/Alert/Alert.js
+++ b/src/Alert/Alert.js
@@ -7,7 +7,7 @@ import { getComponentVariant, createComponent } from '../utils';
 const StyledAlert = createComponent({
   name: 'Alert',
   style: ({ variant, theme }) => {
-    const { backgroundColor, fontColor } = getComponentVariant(theme, 'Alert', variant);
+    const variantStyles = getComponentVariant(theme, 'Alert', variant);
 
     return css`
       padding: 1rem;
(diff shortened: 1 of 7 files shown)


src/Alert/Alert.js
```diff
@@ -7,24 +7,22 @@ import { getComponentVariant, createComponent } from '../utils';
...
+    const variantStyles = getComponentVariant(theme, 'Alert', variant);
 
     return css`
       padding: 1rem;
       margin-bottom: 1rem;
       border: 0;
       font-size: 14px;
       font-family: ${theme.typography.fontFamily || 'inherit'};
-      border-radius: ${theme.elementRadius}px;
-      border-left: 4px solid ${fontColor};
```
cehsu: Did we decide to remove the border-left on alerts? Sorry if I missed this thread.

PR #68, opened 2019-08-23
feat(Dropdown): Style updates

mathewmorris:


src/Dropdown/Dropdown.js
```diff
@@ -346,7 +375,12 @@ Dropdown.Item = function DropdownItem({ closeOnClick = true, onClick, ...props }
       onClick();
     }
   };
-  return <StyledDropdownItem onClick={handleClick} {...props} />;
+  return (
+    <StyledDropdownItem onClick={handleClick} icon={icon} iconSize={iconSize} {...props}>
```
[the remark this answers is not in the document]
mathewmorris: I think that sounds like a good idea. Maybe we do that in another PR?

src/Dropdown/Dropdown.js
```diff
@@ -307,37 +316,50 @@ const StyledDropdownItem = createComponent({
...
-    padding: 0.25rem 1rem;
+    margin: 0;
+    padding: 8px;
     transition: 125ms background;
     outline: none;
     appearance: none;
     border: 0;
+    border-radius: ${theme.radius}px;
     font: inherit;
+    font-size: 14px;
```
erikshestopal: Can we add `font-weight: 500` per the Figma specs since the text seems a little thin in the deploy preview?

src/Dropdown/Dropdown.js
```diff
@@ -285,8 +284,18 @@ Dropdown.Header = ({ title, children }) => (
...
+Dropdown.Section = createComponent({
+  name: 'DropdownSection',
+  as: Flex,
+  style: ({ theme }) => css`
+    padding: 16px 8px 8px 8px;
+    flex-direction: column;
+
+    &:not(:last-of-type) {
+      border-bottom: 1px solid ${theme.colors.greyLight};
+    }
```
erikshestopal: @mathewmorris What necessitated the addition of this `Section` component? It seems to me like `Dropdown.Body` had the necessary padding already, which was removed. Can we stick as close the original implementation as possible and only add/remove things surgically to match the new Figma requirements? That would be greatly appreciated! 😄 💪 👍

mathewmorris: What do you recommend we do with the separators? Right now they are spanning across 100% of the dropdown using the `Dropdown.Section`. I'm not sure how keeping the padding in the body would allow that to happen.
That was the main reason I added `Dropdown.Section`.

erikshestopal: @mathewmorris I don't see a separator in the molekule preview for dropdown sections... 🤔

kylealwyn: @mathewmorris @erikshestopal I'm also not seeing the icon usage in the deploy preview, but I see it in the github diff...

kylealwyn:
@mathewmorris Looks good so far. The markup in the stories is getting pretty verbose, even for a simple dropdown. I'm curious if we can get to something like the following:
```jsx
<Dropdown trigger={Button}>
  <Dropdown.Title>Yo</Dropdown.Title>
  <Dropdown.Item>One</Dropdown.Item>
  <Dropdown.Item>Two</Dropdown.Item>
  <Dropdown.Divider />
  <Dropdown.Item color="red">Sign out</Dropdown.Item>
</Dropdown>
```

Considered changes:
- No more Header/Footer elements. Not sure if we need them anymore
- Render Dropdown children directly into padded container. The Divider element will need negative margins to be flush with the edges

@erikshestopal @cehsu

src/Dropdown/Dropdown.js
mathewmorris: Yeah, I'm not sure what's going on with the preview. I'm also seeing everything locally.

kylealwyn: I'm seeing the changes updated now on the preview

PR #69, opened 2019-08-26
feat: Add Storybook and component stories

kylealwyn:
## Summary
- Migrate to Storybook for documentation
- Update to `@testing-library/react`


cehsu:
Looks awesome 🎉 
The a11y add on is sweet. What are your thoughts on including this addon: https://github.com/vertexbz/storybook-addon-react-live-edit?

kylealwyn:
@cehsu 
- Added theme with our logo
- Live edit didn't work out of the box so something we can maybe look at adding later
- Don't see a huge need for actions but there if we need it

My preference is to get what we have across the finish line and get v5 out

kylealwyn:
@erikshestopal See above comments. @cehsu suggested a plugin that didn't work immediately OOTB so tabled it for now. 

I am curious though, how often/do you use the live edit - I only ask because it seems useful in theory but I've never found myself actually using it.

erikshestopal:
@kylealwyn ok, you make a good point since I've only really used it to QA `molekule` updates. Otherwise, I don't use it 😄

PR #71, opened 2019-09-06
Switch markup cleanup

kylealwyn:
The updated styles to the Switch look awesome - the legacy markup semantics were not. Cleaned those up while maintaining the slick UX @mathewmorris introduced

### DIFF
--- a/src/Form/Switch.js
+++ b/src/Form/Switch.js
@@ -88,14 +88,12 @@ export class Switch extends React.Component {
 
   handleChange = () => {
     const { onChange, name } = this.props;
-    const { value } = this.state;
-    const nextValue = !value;
 
-    this.setState({ value: nextValue }, () => {
-      if (typeof onChange === 'function') {
-        onChange(name, nextValue);
-      }
-    });
+    if (typeof onChange === 'function') {
+      onChange(name, !this.state.value);
+    } else {
+      this.setState(state => ({ value: !state.value }));
+    }
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


src/Form/Switch.js
```diff
@@ -1,120 +1,116 @@
...
+  as: 'label',
+  style: ({ theme, trackColor, thumbSize, trackInset, value }) => css`
+    position: relative;
+    display: inline-block;
+    cursor: pointer;
+    width: ${thumbSize * 2}px;
+    height: ${thumbSize}px;
+    background: ${value ? theme.colors[trackColor] : 'transparent'};
+    border: ${trackInset}px solid ${value ? theme.colors[trackColor] : theme.colors.grey};
+    border-radius: ${thumbSize * 2}px;
```
mathewmorris: Maybe add a little fade to those colors?

```suggestion
    border-radius: ${thumbSize * 2}px;
    transition: background-color 0.3s, border-color 0.3s;
```

kylealwyn: Nice

src/Form/Switch.js
```diff
@@ -1,120 +1,117 @@
...
-  update(newProps) {
-    this.setState({ on: newProps.value });
+    return null;
   }
 
   handleChange = () => {
     const { onChange, name } = this.props;
-    const { on } = this.state;
+    const { value } = this.state;
+    const nextValue = !value;
```
erikshestopal: Thoughts on using a state updater function to simplify this a bit? 
```js
const handleChange = () => {
  const { onChange, name } = this.props;

  this.setState(
    state => ({ value: !state.value }),
    () => {
      if (typeof onChange === 'function') {
        onChange(name, this.state.value);
      }
    }
  );
};
```

PR #73, opened 2019-09-10
feat: Add floating label to select

cehsu:
These changes bring the select component up to date with other form elements that support floating labels.

Screen Shot 2019-09-10 at 10 01 34 AM

Task: https://app.asana.com/0/1127722639002160/1135318045362828/f


kylealwyn:
Could we have this go into v5, which will be landing soon,  so we don’t have to worry about any conflicts?

src/Form/Select.js
```diff
@@ -112,19 +116,25 @@ class Select extends Component {
   };
 
   render() {
-    const { id, name, options, placeholder, error, size = 'md', label, ...props } = this.props;
+    const { id, name, options, placeholder, error, size = 'md', label, floating, ...props } = this.props;
     const { value } = this.state;
+    const isFloating = floating && value != undefined && `${value}`.trim();
+    const StyledLabel = (
```
schlegz: nit: rename this to `FloatingLabel` so that is more obvious what this component is.

src/Form/Select.js
```diff
@@ -112,19 +116,25 @@ class Select extends Component {
   };
 
   render() {
-    const { id, name, options, placeholder, error, size = 'md', label, ...props } = this.props;
+    const { id, name, options, placeholder, error, size = 'md', label, floating, ...props } = this.props;
     const { value } = this.state;
+    const isFloating = floating && value != undefined && `${value}`.trim();
```
choochootrain: is `!= undefined` intentional? i've purged my brain of js quirk knowledge for my sanity so ignore if this is me being dumb

cehsu: @kylealwyn introduced this clever hack to me, we only want to disregard null and undefined, 0 and false should still count. We do the same thing for isFloating in Input.js.

PR #79, opened 2019-09-15
feat(Text): New Text component; upgrade to styled-system@v5

kylealwyn:
### New `<Text />` component
- `<Text as="h1" fontSize={100} color="primaryLight" />`
- Should allow us to remove many of our one-off styled components, which are typically used only to style some text

### Upgrade styled-system
- Better performance for foundational components like Box and Flex with `compose`
- Better support for responsive styles, e.g. `<Box m={{ _: 0, sm: 12 }} />`, which says for all screens, apply 0 margin, and for small screens and up, apply a 12px margin

### Clean up theme.js
- Better organization
- Remove grid nesting
- Add 24px to space scale
- Allow breakpoints to be overridden

### Clean up `<Column />`
- Remove `order` style so styled-system can apply `order` responsively, e.g. `<Column order={{ _: -1, sm: 1 }} />`
- Prettier wasn't liking some of the implicit ternary returns

### Fix Dropdown stories
- Storybook was not liking hooks in the base story
- Modal was using old Dropdown syntax

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,6 +1,7 @@
 import babel from 'rollup-plugin-babel';
 import filesize from 'rollup-plugin-filesize';
 import cjs from 'rollup-plugin-commonjs';
+import resolve from 'rollup-plugin-node-resolve';
 import pkg from './package.json';
 
 export default {
@@ -15,10 +16,17 @@ export default {
       format: 'es',
     },
   ],
-  external: [...Object.keys(pkg.dependencies || {}), ...Object.keys(pkg.peerDependencies || {})],
+  external: [
+    ...Object.keys(pkg.dependencies || {}).filter(dep => dep !== 'styled-components'),
+    ...Object.keys(pkg.peerDependencies || {}),
+  ],
   plugins: [
     babel({
       babelrc: true,
+    }),
+    resolve({
+      only: ['styled-components'],
+      dedupe: ['react', 'react-dom', 'styled-components'],
     }),
     cjs({
       include: 'node_modules/**',
--- a/src/Alert/Alert.js
+++ b/src/Alert/Alert.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from 'styled-components';
 import { space } from 'styled-system';
+import propTypes from '@styled-system/prop-types';
 import { getComponentVariant, createComponent } from '../utils';
 
 const StyledAlert = createComponent({
@@ -29,10 +30,11 @@ const StyledAlert = createComponent({
 });
 
 /** Alerts are typically used to display meaningful copy to users - typically notifying the user of an important message. */
-const Alert = props => <StyledAlert {...props} />;
+const Alert = React.forwardRef((props, ref) => <StyledAlert {...props} ref={ref} />);
 
 Alert.propTypes = {
   variant: PropTypes.string,
+  ...propTypes.space,
 };
 
 Alert.defaultProps = {
--- a/src/Badge/Badge.js
+++ b/src/Badge/Badge.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css } from 'styled-components';
 import { space } from 'styled-system';
+import propTypes from '@styled-system/prop-types';
 import { getComponentVariant, createComponent, getComponentSize } from '../utils';
 
 const StyledBadge = createComponent({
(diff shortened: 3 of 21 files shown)


rollup.config.js
```diff
@@ -15,11 +16,18 @@ export default {
       format: 'es',
     },
   ],
-  external: [...Object.keys(pkg.dependencies || {}), ...Object.keys(pkg.peerDependencies || {})],
+  external: [
+    ...Object.keys(pkg.dependencies || {}).filter(dep => dep !== 'styled-components'),
```
cehsu: Would it be worthwhile to add a comment about why we're doing this?

src/Alert/Alert.js
```diff
@@ -29,10 +30,11 @@ const StyledAlert = createComponent({
 });
 
 /** Alerts are typically used to display meaningful copy to users - typically notifying the user of an important message. */
-const Alert = props => <StyledAlert {...props} />;
+const Alert = React.forwardRef((props, ref) => <StyledAlert {...props} ref={ref} />);
```
cehsu: 🙏

src/Badge/Badge.js
```diff
@@ -22,11 +23,12 @@ const StyledBadge = createComponent({
   },
 });
 
-const Badge = props => <StyledBadge {...props} />;
+const Badge = React.forwardRef((props, ref) => <StyledBadge {...props} ref={ref} />);
```
cehsu: 🙏

src/Alert/Alert.js
```diff
@@ -29,10 +30,11 @@ const StyledAlert = createComponent({
...
 
 Alert.propTypes = {
   variant: PropTypes.string,
+  ...propTypes.space,
```
cehsu: 😎

src/Text/Text.js
```diff
@@ -0,0 +1,24 @@
...
+import { createComponent } from '../utils';
+
+const StyledText = createComponent({
+  name: 'Text',
+  tag: 'span',
+  style: compose(
+    space,
+    color,
+    typography
+  ),
```
cehsu: So slick 😎

src/Text/Text.spec.js
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  test('"as" prop', () => {
+    const { asFragment } = renderWithTheme(<Text as="h1">Hey!</Text>);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('"color" prop', () => {
+    const { asFragment } = renderWithTheme(<Text color="red">Hey!</Text>);
+    expect(asFragment()).toMatchSnapshot();
```
cehsu: While we're at it, should we add a `space` prop test as well?

src/Text/Text.stories.js
```diff
@@ -0,0 +1,15 @@
...
+export default {
+  title: 'Components|Text',
+  component: Text,
+};
+
+export const Basic = () => (
+  <>
+    <Text>The basic text component renders as a {'<span />'}</Text>
+    <Text as="h1">It can render as other tags, like this {'<H1/>'}</Text>
+    <Text color="red">It can use colors from the theme</Text>
```
cehsu: Similar to above, should we add a spacing prop case while we're going to the trouble to illustrate these props, or is it similar enough to the existing spacing props on other components, that there's no need for the additional explanation?

src/Modal/Modal.dropdownExample.js
```diff
@@ -27,16 +27,10 @@ export default class ModalDropdownDemo extends React.Component {
     return (
       <div>
         <Dropdown placement="top" width={250} trigger={<Button variant="success">Open Dropdown</Button>}>
-          <Dropdown.Header title="Dropdown" />
```
cehsu: No more need for the multi section dropdown? Do we want to use the divider here? I think we had added the dropdown to facilitate keyboard accessibility testing.

src/Grid/Column.js
```diff
@@ -32,24 +34,26 @@ const getOffset = (offset, columns) => css`
...
+          ${props[`${breakpoint}Offset`] && getOffset(props[`${breakpoint}Offset`], theme.gridColumns)};
+        `;
+      }
+
+      return css`
+        @media (min-width: ${theme.breakpoints[breakpoint]}) {
+          ${getColumnWidth(props[breakpoint], theme.gridColumns)};
+          ${props[`${breakpoint}Offset`] !== undefined && getOffset(props[`${breakpoint}Offset`], theme.gridColumns)};
+        }
+      `;
```
cehsu: Could we use breakpoint aliases to generate these media queries?

kylealwyn: I'd rather not touch the implementation in this PR

src/Modal/Modal.dropdownExample.js
kylealwyn: wut

src/Text/Text.stories.js
kylealwyn: The latter

PR #81, opened 2019-09-18
Various UI updates from web-app review

kylealwyn:
Going through web-app with fine toothed comb. Various findings will be dropped here.

### DIFF
--- a/src/Form/Input.js
+++ b/src/Form/Input.js
@@ -357,7 +357,7 @@ export class Input extends Component {
 
     const { focused, height, value } = this.state;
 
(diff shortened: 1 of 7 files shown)


src/Form/Input.js
```diff
@@ -357,7 +357,7 @@ export class Input extends Component {
 
     const { focused, height, value } = this.state;
 
-    const isFloating = (floating && value !== undefined && !!`${value}`.trim()) || (floating && focused);
+    const isFloating = (floating && value != null && !!`${value}`.trim()) || (floating && focused);
```
mathewmorris: ah yeah. This is better

PR #82, opened 2019-09-18
Add Draggable Modal

jlemm45:
**Ticket**
https://app.asana.com/0/1127722652598124/1127722950634985/

**Summary**
- Ability to pass `draggable` to modal component and have a modal that can be moved around the viewport
- Ability to pass extra options for Draggable using `draggableOptions`
- Add story example

**Questions**
Should we export `Draggable` as it's own component out of `molekule`? 
I also ensured that the modal have a `title` prop present so a header displays to use this feature per ticket guidelines. Just want to confirm that is the correct functionality.

**Example**
Sep-18-2019 18-34-54

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2346,6 +2346,16 @@
             "@babel/runtime": "^7.4.5"
           }
         },
+        "react-draggable": {
+          "version": "3.3.2",
+          "resolved": "https://registry.npmjs.org/react-draggable/-/react-draggable-3.3.2.tgz",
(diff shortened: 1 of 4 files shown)


kylealwyn:
@jlemm45 Thanks for setting this up. After giving this some more thought, this is something that we should probably land in the EMR rather than molekule core. We won't be using draggables on our consumer apps so we shouldn't bloat those bundles unnecessarily

jlemm45:
@kylealwyn That makes sense as more of a doctor only feature. I can transition this into a PR over there.

schlegz:
@jlemm45 Based on your previous comment, I'm assuming this PR is now invalid and you will move this work over to `web-app`?

PR #83, opened 2019-10-15
fix(Modal): Opens long-content modal at top

mathewmorris:
This was introduced by `react-focus-on` & it's predecessor `react-focus-lock`. It wasn't necessarily a bug since it was doing what it was intended to do, and that is to make the modal accessible. In doing so, you lock the focus into the modal, and the first item it would focus on is an input or button. This would cause the modal to open near the bottom of the scroll length. I needed to make the header accessible as well since you could not navigate to the close button via the keyboard alone.

- Adds aria roles and tabIndex to Modal Header and buttons.

## Before
modal-middle

## After
modal_top

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -102,7 +102,7 @@ export function Modal({ children, title, animationDuration, showClose, onClose, 
           {state => (
             <FocusOn onEscapeKey={handleClose} enabled={isOpen}>
               <Backdrop transitionState={state} onClick={handleBackdropClick}>
-                <ModalContent transitionState={state} onClick={handleContentClick} {...props}>
+                <ModalContent transitionState={state} onClick={handleContentClick} aria-modal="true" {...props}>
                   {title && <Modal.Header title={title} showClose={showClose} />}
                   {children}
                 </ModalContent>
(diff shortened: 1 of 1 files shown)


kylealwyn:
This didn't fix the issue for me on the deploy preview

mathewmorris:
I have attached a gif of before/after my changes. @kylealwyn Would you mind showing a gif or screenshot of what you're seeing?

kylealwyn:
modal-long

let me know if I'm doing something wrong

mathewmorris:
Hmmmm..... That's strange. It only happens when I make my window very small. I guess that's what I get for working on a huge 4k monitor lol.

kylealwyn:
@mathewmorris ping here

mathewmorris:
@kylealwyn Try this out again. I think I finally figured out the culprit. Should always be at the top of long content no matter the screen size.

src/Modal/Modal.js
```diff
@@ -98,11 +103,11 @@ export function Modal({ children, title, animationDuration, showClose, onClose,
   return (
     <ModalContext.Provider value={{ handleClose }}>
       <Portal>
-        <Transition in={isOpen} timeout={animationDuration}>
+        <Transition in={isOpen} timeout={animationDuration} onEntering={scrollToTop}>
           {state => (
             <FocusOn onEscapeKey={handleClose} enabled={isOpen}>
```
kylealwyn: @mathewmorris I think Escape is already supported here - was that not the case? Looking at the new escape support in the Header

kylealwyn:
Screen Shot 2019-11-27 at 3 07 31 PM

@mathewmorris Working here. Let's drop the outline when the title is focused but other than that, looks great!

src/Modal/Modal.js
mathewmorris: Oh that's for the close Icon if `Enter` is pressed while focused on the icon.

PR #85, opened 2019-11-07
feat: Allow Accordion items to be open by default

erikshestopal:
## Summary 
A recent requirement came up with the patient notification update work that specifies that an accordion item needs to be open by default is a certain condition is met. So, for example, in the screenshot below, we want to keep the accordion item open only if the patient has at least one notification type enabled. 

Screen Shot 2019-11-07 at 10 55 15 AM

### DIFF
--- a/src/Accordion/Accordion.js
+++ b/src/Accordion/Accordion.js
@@ -12,6 +12,7 @@ const AccordionItemProps = {
   content: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
   renderHeader: PropTypes.func,
   renderContent: PropTypes.func,
+  isOpenByDefault: PropTypes.bool,
 };
 
 const AccordionContainer = createComponent({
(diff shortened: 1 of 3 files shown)


src/Accordion/Accordion.js
```diff
@@ -99,9 +100,15 @@ export default class Accordion extends Component {
...
-  state = {
-    openList: [],
-  };
+  constructor(props) {
+    super(props);
+
+    const openByDefault = props.items.map((item, index) => (item.isOpenByDefault ? index : undefined));
+
+    this.state = {
+      openList: openByDefault.filter(e => e !== undefined),
```
choochootrain:
```suggestion
      openList: props.items.filter(item => item.isOpenByDefault),
```

erikshestopal: @choochootrain `#handleItemToggle` relies on an item's index to show/hide items so I don't think this suggestion would work.

erikshestopal:
Closing this PR because it was decided that we won't utilize default open functionality for the patient notification updates.

Issue #88, opened 2019-11-11
chore(Lodash): Performance improvements


mathewmorris:
The entire lodash library is being pulled in with molekule. We should only pull in the specific methods that we need for the library.

PR #89, opened 2019-11-11
chore: Uses specific lodash methods

mathewmorris:
In reference to #88 

This should only pull in certain methods needed from lodash instead of the entire library.

### DIFF
--- a/src/Form/FormError.js
+++ b/src/Form/FormError.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-nested-ternary */
 import React, { useContext } from 'react';
 import { css } from 'styled-components';
-import { get } from 'lodash';
+import get from 'lodash/get';
 import { createComponent } from '../utils';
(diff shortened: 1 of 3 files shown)


erikshestopal:
@mathewmorris does webpack not tree shake all the code that's not used?

mathewmorris:
> @mathewmorris does webpack not tree shake all the code that's not used?

I believe molekule uses babel to build. So, I'm not sure that babel actually treeshakes outside of it's own methods from core-js and other plugins.

mathewmorris:
Though, Rollup might do the treeshaking for us. I'll confirm. I'll also analyze the bundles for some concrete evidence.

mathewmorris:
Ah. Rollup is not being used in the build process.

mathewmorris:
@erikshestopal Yeah, so babel is just compiling everything into the lib folder. Nothing is being bundled.

PR #94, opened 2019-12-17
fix(Modal): Safeguards against onEntering error

mathewmorris:
https://sentry.io/organizations/heydoctor/issues/1393699250/?referrer=slack

The modal backdrop would sometimes render after `onEntering` was called, so I wrapped in a conditional to prevent it from calling `.scroll` on the null object.

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -91,7 +91,12 @@ export function Modal({ children, title, animationDuration, showClose, onClose, 
   };
 
   const scrollToTop = () => {
-    modalRef.current.scroll(0, 0);
+    // short delay to make sure backdrop is rendered
+    setTimeout(() => {
+      if (modalRef.current) {
+        modalRef.current.scroll(0, 0);
+      }
+    }, 1);
   };
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/Modal/Modal.js
```diff
@@ -91,7 +91,12 @@ export function Modal({ children, title, animationDuration, showClose, onClose,
   };
 
   const scrollToTop = () => {
-    modalRef.current.scroll(0, 0);
+    // short delay to make sure backdrop is rendered
+    setTimeout(() => {
+      if (modalRef.current) {
+        modalRef.current.scroll(0, 0);
+      }
+    }, 1);
```
schlegz: This seems like an assumption that all rendering needs just a millisecond, which may not be true if a browser/phone/pc is very busy.

mathewmorris: That's true. Do we want to just keep the tick out? I think the safeguard against the ref possibly not being set is okay. Though, it's more of a scapegoat than anything.

mathewmorris:
> I think the conditional check is fine, not sure if there is still a race condition that needs to be dealt with here but it seems the main problem is the possible null value. @erikshestopal @vietnogigoodrx thoughts?

@erikshestopal @vietnogigoodrx Do you have any input here? Otherwise, I think we're good to mrege. Let me know! 👍

PR #95, opened 2020-01-08
feat(Modal): Only renders modal in DOM when visible

mathewmorris:
This is specific to accessibility and testing in general (though it's not a change to make testing work right, this was a problem that was unearthed by testing actually).

Modals and their content were being rendered when not visible, and so readable by screen readers which would not make sense if it's not showing. Now, the modal and it's content will render when it is supposed to be visible. 🎉 

Makes the close icon a button. Also, it gives autofocus back to content.

No breaking changes.

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -6,8 +6,8 @@ import { Transition } from 'react-transition-group';
 import { FocusOn } from 'react-focus-on';
 import Portal from '../Portal';
 import Flex from '../Flex';
-import Box from '../Box';
 import Icon from '../Icon';
+import Button from '../Button';
 import { createComponent, themeGet } from '../utils';
 
 const ModalContext = createContext({});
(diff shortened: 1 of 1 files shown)


kylealwyn:
@mathewmorris Awesome! One note - It looks like the inputs in the Modals have the autoFocus attribute but they're not autofocusing

mathewmorris:
@kylealwyn Yeah, looks like the focus goes to the Cancel button first (which doesn't have a nice focus state I might add). I will get the autofocus on the input 👍

src/Modal/Modal.js
```diff
@@ -185,27 +185,20 @@ Modal.Header = ({ title, children, showClose = true }) => {
     <ModalHeader>
       <ModalHeaderInner>
         <Flex alignItems="center">
-          {title && (
-            <Modal.Title role="heading" tabIndex="0">
-              {title}
-            </Modal.Title>
-          )}
+          {title && <Modal.Title role="heading">{title}</Modal.Title>}
```
vietnogigoodrx: intentional to remove the tab-index?

src/Modal/Modal.js
```diff
@@ -185,27 +185,20 @@ Modal.Header = ({ title, children, showClose = true }) => {
...
-                style={{ cursor: 'pointer' }}
-                size={24}
-                onClick={handleClose}
-                onKeyDown={handleKeyDown}
-                role="button"
-                aria-label="Close Modal"
-                tabIndex="0"
-              />
-            </Box>
+            <Button
```
vietnogigoodrx: I like to specify `type="button"` incase this component is wrapped into a form.

src/Modal/Modal.js
mathewmorris: Yes. So it doesn't take the focus when the content of the modal has something that needs autofocus.

PR #99, opened 2020-01-25
fix(Checkbox/Radio): Fixes regression from 5.1.1 update

mathewmorris:
Coming from feedback in https://github.com/heydoctor/web-app/pull/408

This will solve the regressions from that update while still keeping the features requested.

### DIFF
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -150,8 +150,10 @@ const CheckboxContainer = createComponent({
 
       ${isHorizontal &&
         css`
-          & + & {
-            margin-left: 18px;
+          @media (min-width: ${theme.breakpoints.sm}) {
+            & + & {
+              margin-left: 18px;
+            }
           }
         `}
 
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -139,7 +139,7 @@ const CheckboxContainer = createComponent({
         !disabled &&
         css`
           ${CheckboxIcon} {
-            color: ${theme.colors.greyDarker};
+            border-color: ${theme.colors.greyDarker};
           }
         `}
     }
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -221,7 +221,7 @@ export class Checkbox extends React.Component {
     onChange() {},
     disabled: false,
     styles: {},
-    label: 'Checkbox',
+    label: null,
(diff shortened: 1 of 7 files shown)


mathewmorris:
@heydoctor/engineering Could I get a couple eyes on this when you get the chance?

src/Form/Checkbox.js
```diff
@@ -111,47 +115,80 @@ const CheckboxLabel = createComponent({
...
-      margin-left: ${horizontal ? 12 : 0}px;
-      margin-top: ${horizontal ? 0 : 4}px;
-    }
+    return css`
+    display: inline-flex;
+    align-items: flex-start;
+    justify-content: flex-start;
+    margin-bottom: 4px;
+    user-select: none;
+    cursor: pointer;
```
erikshestopal: stylistic nit: can we indent these styles here and below 2 spaces to the right to be consistent with other styles? 

```js
return css`
  display: inline-flex;
  // etc 
`
```

src/Form/Checkbox.stories.js
```diff
@@ -1,52 +1,39 @@
...
 
 export default {
   title: 'Components|Forms/Checkbox',
   component: Checkbox,
 };
 
-export const Basic = () => (
-  <Checkbox id="checkbox" name="checkbox" label="I'm a checkbox" disabled={boolean('Disabled', false)} />
-);
+export const Basic = () => <Checkbox />;
```
erikshestopal: Can we add an example for a checkbox with the `isRadio` prop?

vietnogigoodrx:
Should we make this wrap/responsive?

Screen Shot 2020-01-29 at 9 45 27 AM

mathewmorris:
@erikshestopal Wanna give this another pass before merging?

src/Form/Checkbox.stories.js
```diff
@@ -12,6 +12,8 @@ export const Basic = () => <Checkbox />;
 
 export const Label = () => <Checkbox label="I'm a checkbox" disabled={boolean('Disabled', false)} />;
 
+export const isRadio = () => <Checkbox isRadio />;
```
erikshestopal: I'm sorry, super nit: 
```suggestion
export const Radio = () => <Checkbox isRadio />;
```

mathewmorris: 😆 No worries!

PR #100, opened 2020-01-29
feat: OrderedList

vietnogigoodrx:
### 📃 Summary
Customized ordered list requested by the design team.

Screen Shot 2020-01-29 at 2 33 35 PM

### 🏷 Asana Task
https://app.asana.com/0/0/1159034716999539/f
### 📸 Screenshots
Screen Shot 2020-01-29 at 2 35 08 PM

### 🧪 QA
See preview
	
### 📈 Risk Analysis
✅ Does this affect patients?
✅ Does this affect clinical?
✅ Does this mutate data?
✅ Does it touch PHI?
✅ Does it touch payments?

### DIFF
--- a/src/Lists/OrderedList.js
+++ b/src/Lists/OrderedList.js
@@ -19,7 +19,7 @@ const StyledOrderedList = createComponent({
       position: relative;
       margin-top: 18px;
 
-      &:first-child {
+      &:first-of-type {
         margin-top: 0;
       }
 
--- a/src/Lists/OrderedList.js
+++ b/src/Lists/OrderedList.js
@@ -1,13 +1,13 @@
-import React from "react";
-import { css } from "styled-components";
-import Box from "../Box";
-import { createComponent } from "../utils";
+import React from 'react';
+import { css } from 'styled-components';
+import Box from '../Box';
+import { createComponent } from '../utils';
 
 const bubbleSize = 22;
 
 const StyledOrderedList = createComponent({
-  name: "OrderedList",
-  as: "ol",
+  name: 'OrderedList',
+  as: 'ol',
   style: ({ theme }) => css`
     counter-reset: my-counter;
     list-style: none;
@@ -25,7 +25,7 @@ const StyledOrderedList = createComponent({
 
       &::before {
         content: counter(my-counter);
-        font-family: ${theme.typography.headerFontFamily || "inherit"};
+        font-family: ${theme.typography.headerFontFamily || 'inherit'};
         position: absolute;
         left: ${bubbleSize * -1 - theme.space[2]}px;
         line-height: ${bubbleSize}px;
@@ -42,11 +42,9 @@ const StyledOrderedList = createComponent({
 
 /** An ordered list that has customize styling
  */
-const OrderedList = React.forwardRef((props, ref) => (
-  <StyledOrderedList {...props} ref={ref} />
-));
+const OrderedList = React.forwardRef(StyledOrderedList);
 
-OrderedList.displayName = "OrderedList";
+OrderedList.displayName = 'OrderedList';
 
 OrderedList.propTypes = {
   ...Box.propTypes
--- a/src/Lists/OrderedList.js
+++ b/src/Lists/OrderedList.js
@@ -1,0 +1,55 @@
+import React from "react";
+import { css } from "styled-components";
+import Box from "../Box";
+import { createComponent } from "../utils";
+
+const bubbleSize = 22;
+
+const StyledOrderedList = createComponent({
+  name: "OrderedList",
+  as: "ol",
+  style: ({ theme }) => css`
+    counter-reset: my-counter;
+    list-style: none;
+    padding: 2px; // Provide space for bubble
+    padding-left: ${bubbleSize + theme.space[2]}px;
+
+    > li {
+      counter-increment: my-counter;
+      position: relative;
+      margin-top: 18px;
+
+      &:first-child {
+        margin-top: 0;
+      }
+
+      &::before {
+        content: counter(my-counter);
+        font-family: ${theme.typography.headerFontFamily || "inherit"};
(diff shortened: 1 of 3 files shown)


src/Lists/OrderedList.js
```diff
@@ -0,0 +1,55 @@
+import React from "react";
```
erikshestopal: The convention in our codebase is to use single quotes for everything except string prop values. Can we please change?

src/Lists/OrderedList.js
```diff
@@ -0,0 +1,55 @@
...
+        text-align: center;
+        border: 1px solid ${theme.colors.grey};
+      }
+    }
+  `
+});
+
+/** An ordered list that has customize styling
+ */
+const OrderedList = React.forwardRef((props, ref) => (
```
erikshestopal: Would it be incorrect to suggest that this is functionally equivalent? 
```suggestion
const OrderedList = React.forwardRef(StyledOrderedList); 
```

vietnogigoodrx: Got an error when doing this: `Warning: forwardRef requires a render function but was given object.`

erikshestopal: I guess it doesn't work with styled-components 🤷‍♂? 😄

src/Lists/OrderedList.js
```diff
@@ -0,0 +1,55 @@
...
+
+const bubbleSize = 22;
+
+const StyledOrderedList = createComponent({
+  name: 'OrderedList',
+  as: 'ol',
+  style: ({ theme }) => css`
+    counter-reset: my-counter;
+    list-style: none;
+    padding: 2px; // Provide space for bubble
```
kylealwyn: Not sure if this matters but this is invalid comment syntax for css

src/Lists/OrderedList.js
```diff
@@ -0,0 +1,55 @@
...
+        top: -2px;
+        border-radius: 50%;
+        text-align: center;
+        border: 1px solid ${theme.colors.grey};
+      }
+    }
+  `
+});
+
+/** An ordered list that has customize styling
```
kylealwyn: Seems unnecessary

src/Lists/OrderedList.stories.js
```diff
@@ -0,0 +1,18 @@
...
+import OrderedList from './OrderedList';
+
+export default {
+  title: 'Components|OrderedList',
+  component: OrderedList
+};
+
+export const Basic = () => (
+  <OrderedList>
+    <li>Apple</li>
```
kylealwyn: Only a suggestion but since we're forced to use li tags based on the styling, making using them somewhat superfluous, a similar pattern we use in molekule is an items prop:
```js
<OrderedList
  items={[
    'Apple',
    'Dog',
    'Groot'
  ]}
/>
```

vietnogigoodrx: So if we use an `items` props then we should allow, JSX elements in addition to strings.  If we go that route, then we we can't utilize JSX children pattern.  Should I just use the children prop?

erikshestopal: @vietnogigoodrx good case for a render prop? 😄 
```jsx
<OrderedList
  renderItem={item => <SomeComponent>{item}</SomeComponent>}
  items={[
    'Apple',
    'Dog',
    'Groot'
  ]}
/>
```

vietnogigoodrx: That would work, but it's restrictive on `SomeComponent`.  I think im going with the children route since its the least complex and it maintains the parent children pattern.

erikshestopal: @vietnogigoodrx what do you mean by restrictive?

vietnogigoodrx: actually, im wrong, its not restrictive since you can add logic in the function

vietnogigoodrx: @erikshestopal is there any advantage vs

```
<OrderedList>
  <SomeComponent>Apple</SomeComponent>
  <SomeComponent>Dog</SomeComponent>
  <SomeComponent>Groot</SomeComponent>
</OrderedList>
```?

erikshestopal: There's less code to write since the only variable that's changing is the content and not the component that is rendering the content. 

The convention in our codebase is to use some sort of `items` array with some sort of `renderItem` prop. 

It seems like you don't agree that this is the best pattern to use here.

vietnogigoodrx: Yea, I dont agree, but it's not a strong disagreement. I will go ahead and add an `items` prop and we can add a `renderItem` prop if we need to in the future.

kylealwyn: @vietnogigoodrx Would you mind elaborating on why you disagree? The render prop pattern is a widely adopted and trusted pattern by both React core teams and community: https://facebook.github.io/react-native/docs/flatlist#renderitem but I'd love to hear your perspective.

Regardless, we haven't seen how else this component can or should be used so we are likely prematurely optimizing here. Maybe there will only ever be strings as children 🤷‍♀. I'll reiterate my suggestion above but will leave to y'all for what and when to add to molekule:
> ... wonder if a general rule of thumb for adding to molekule should be for design to use it twice.

vietnogigoodrx: I disagree because the render prop pattern feels like an overkill for the component.  I am comparing it with the `ol`/`ul` components and both of them are quite simple.  They expect an `li` children that are styled.  I was hoping to maintain this pattern for this component.

The reason I am adding it to to the molekule repo based on Devon's suggestion.  She said that we would be using this design in many places.

kylealwyn: Understood and sounds good regarding intention. I left a couple of other comments but fine with this usage pattern

PR #101, opened 2020-01-31
fix(Modal): Add guard to scrollToTop

choochootrain:
https://sentry.io/organizations/heydoctor/issues/1397040362/?project=290802&query=is%3Aunresolved+scroll&statsPeriod=14d

#cleanupsentry

### DIFF
--- a/src/Modal/Modal.js
+++ b/src/Modal/Modal.js
@@ -91,7 +91,7 @@ export function Modal({ children, title, animationDuration, showClose, onClose, 
   };
 
   const scrollToTop = () => {
-    if (modalRef.current) {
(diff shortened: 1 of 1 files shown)


src/Modal/Modal.js
```diff
@@ -91,7 +91,7 @@ export function Modal({ children, title, animationDuration, showClose, onClose,
   };
 
   const scrollToTop = () => {
-    if (modalRef.current) {
+    if (modalRef.current && modalRef.current.scroll) {
```
erikshestopal: do we have optional chaining in `molekule`?

choochootrain: not at the moment

PR #102, opened 2020-02-05
fix(Button): Rename internal loading prop to isLoading to remove console error

kylealwyn:
Tired of seeing this console error
Screen Shot 2020-02-04 at 4 33 46 PM

### DIFF
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -52,7 +52,7 @@ const StyledButton = createComponent({
     theme,
     block,
     disabled,
-    loading,
+    isLoading,
     borderRadius = theme.radius,
     colorFocus = theme.colors.colorFocus,
   }) => {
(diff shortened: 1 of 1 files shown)


src/Button/Button.js
```diff
@@ -52,7 +52,7 @@ const StyledButton = createComponent({
     theme,
     block,
     disabled,
-    loading,
+    isLoading,
```
vietnogigoodrx: +1

src/Button/Button.js
```diff
@@ -133,13 +133,14 @@ const renderIcon = (icon, props) => <Icon name={icon} {...props} />;
...
 const Button = React.forwardRef(
-  ({ children, leftIcon, leftIconProps, rightIcon, rightIconProps, colorFocus, ...rest }, ref) => (
+  ({ children, leftIcon, leftIconProps, rightIcon, rightIconProps, colorFocus, loading, ...rest }, ref) => (
     <StyledButton
       ref={ref}
       hasText={!!children}
       leftIcon={leftIcon}
       rightIcon={rightIcon}
       colorFocus={colorFocus}
+      isLoading={loading || undefined}
```
vietnogigoodrx: `|| undefined` 🤔Is it because we don't want false?

kylealwyn: Cruft from first take - not necessary

PR #103, opened 2020-03-31
fix success button background color

vietnogigoodrx:


vietnogigoodrx:
I suppose it all boils down to semantics.   Started off with https://app.asana.com/0/1158229319553377/1168988755094203/f,  the problem was that goodrx stylesheet was showing blue for a button with a success variant.  Its suppose to be green, but it is actually blue because the secondary color for goodrx is blue.  The success button looks fine on heydoctory because the secondary button for hey doctor happens to be green.

choochootrain:
can we do this in `web-app` with a theme override? https://github.com/heydoctor/web-app/blob/master/src/config/clients/goodrx.ts#L29

im wary of making these changes globally in molekule where they could regress other usages

vietnogigoodrx:
> can we do this in `web-app` with a theme override? https://github.com/heydoctor/web-app/blob/master/src/config/clients/goodrx.ts#L29
> 
> im wary of making these changes globally in molekule where they could regress other usages

Closing PR in favor of this solution.

PR #108, opened 2020-05-14
chore: add typescript

vietnogigoodrx:
### 📃 Summary
Add minimal settings to allow typescript to run.  Will convert each files and add better typings as time goes on.

### 🧪 QA
Pull down code and run `yarn dev` and everything should compile.  Modify icon.tsx and changes should show up in the browser.  If an error occurs, it should show an error in the browser.  `yarn build` and `yarn test` should continue working.
	
### 📈 Risk Analysis
- [ ] Does this mutate data?
- [ ] Does it touch PHI?
- [ ] Does it touch payments?


.storybook/presets.js
```diff
@@ -1 +0,0 @@
-module.exports = ['@storybook/addon-docs/react/preset'];
```
vietnogigoodrx, line 1: added to main.js

package.json
```diff
@@ -16,8 +16,9 @@
   "scripts": {
     "clean": "rimraf lib",
     "watch": "yarn rollup --watch",
-    "build": "npm run clean && babel src --out-dir lib",
+    "build": "npm run clean && babel src --out-dir lib --extensions '.ts,.tsx,.js,.jsx'",
```
schlegz, line 19: Curious why we include the `jsx` extension here but not in the extensions list in the other config files.

vietnogigoodrx: missed it in the other file, added

vietnogigoodrx:
Can I get somebody to review this PR?

vietnogigoodrx:
> Remove `package-lock.json` since we moved to using `yarn.lock`.

did it in another PR

albertovillalobos:
@vietnogigoodrx answering your quesiton here as well:

The check that says “complete setup” you can ignore because this branch was cut before chromatic changes made it in
The other one looks like a component actually changed
and needs to be approved, it looks like you “created” a new icon (actually just changed extension), So it’s detected as a new change, which i’ll approve

PR #109, opened 2020-05-14
chore: use yarn so that ci will run test like how we would locally

vietnogigoodrx:
### 📃 Summary
Have problems matching snapshots locally vs on CI because CI uses npm and locally we use yarn.  This would allow only one lock file.

### 📈 Risk Analysis
- [ ] Does this mutate data?
- [ ] Does it touch PHI?
- [ ] Does it touch payments?


.travis.yml
```diff
@@ -2,8 +2,8 @@ language: node_js
 node_js:
   - 10
 install:
-  - npm install -g codecov
+  - yarn install -g codecov
```
vietnogigoodrx, line 5: any reason why we used npm in the first place?

kylealwyn:
@vietnogigoodrx this repository used npm initially. you'll need to add a yarn.lock and use `yarn import` if we want to make this change.

albertovillalobos:
@vietnogigoodrx we can ignore the `UI Review - Complete Setup` check on this PR
It's weird that the snapshots are changing tho, they are meant to be deterministic. 🤔

PR #110, opened 2020-05-14
chore: add better types

alextranwork:
Added more typings, will follow up with more typings and publishing the types


src/defaultThemeColors.ts
```diff
@@ -0,0 +1,56 @@
+import { ThemeColors } from './types/ThemeColors';
+
+export const defaultThemeColors: ThemeColors = {
```
vietnogigoodrx, line 3: Took out of createTheme to make it easier to work with.

src/defaultThemeSizes.ts
```diff
@@ -0,0 +1,120 @@
+import { ThemeSizes } from './types';
+
+export const defaultThemeSizes: ThemeSizes = {
```
vietnogigoodrx, line 3: Took out of createTheme to make it easier to work with.

src/createTheme.ts
```diff
@@ -0,0 +1,193 @@
+import { DefaultTheme } from 'styled-components';
+import { ThemeVariants } from './types/ThemeVariants';
+import { defaultThemeColors } from './defaultThemeColors';
+import { ThemeBreakpoints } from './types/ThemeBreakpoints';
+import { ButtonVariants } from './types/ButtonVariants';
+import { BadgeVariants } from './types/BadgeVariants';
+import { defaultThemeSizes } from './defaultThemeSizes';
+
+export const createTheme = (customTheme?: DefaultTheme): DefaultTheme => {
```
vietnogigoodrx, line 9: originally from src/theme.js, converted to ts and gave it a better name and took some contants out to make the file shorter.

src/createTheme.ts
```diff
@@ -0,0 +1,193 @@
+import { DefaultTheme } from 'styled-components';
+import { ThemeVariants } from './types/ThemeVariants';
+import { defaultThemeColors } from './defaultThemeColors';
+import { ThemeBreakpoints } from './types/ThemeBreakpoints';
+import { ButtonVariants } from './types/ButtonVariants';
+import { BadgeVariants } from './types/BadgeVariants';
+import { defaultThemeSizes } from './defaultThemeSizes';
```
erikshestopal, line 7: I can see what you're trying to do here but there is a cleaner way of achieving stronger type safety by using Typescript's `as const` syntax. We can avoid all these intermediate variant types by marking the style object `as const`:

```tsx
export const createTheme = () => {
  const colors = {
    primaryLightest: '#CADCFF',
    primaryLight: '#226EFF',
    primary: '#0958F3',
    primaryDark: '#0046CE',
  } as const;

  const buttonVariants = {
    primary: {
      backgroundColor: colors.primary,
      color: colors.white,
      '&:hover': {
        backgroundColor: colors.primaryLight,
      },
      '&:active': {
        backgroundColor: colors.primaryDark,
      },
      '&:disabled': {
        backgroundColor: colors.primaryLightest,
      },
    },
  } as const;

  return {
    colors,
    space: [0, 4, 8, 16, 24, 32, 64, 126, 256] as const,
    typography: {
      fontSize: 16,
      color: colors.black,
      bodyFontFamily: 'Avenir',
      headerFontFamily: 'Tiempos',
    } as const,

    // ...etc
  };
};

```

Screen Shot 2020-05-22 at 12 06 28 PM

erikshestopal (changes requested):

src/createTheme.ts
vietnogigoodrx: neat! let me update to use `const`

vietnogigoodrx: 🤔 Does that mean we can only ever have one default and the type is strictly based on the default?

erikshestopal: @vietnogigoodrx can you give me an example of you mean?

vietnogigoodrx: While implementing the `as const`, I started feeling that we should have explicit types defined.    I think it's easier to understand the types when they're explicit.  I am also afraid that changes to the default value may have unintentional changes to the type.  I was planning to export these types so that the web-app, so I think its a good idea to be explicit to reduce the chances of breakage.

erikshestopal: @vietnogigoodrx I'm not sure what you mean by explicit types. 

A type of 
```ts
type Colors = {
  default: 'some-hex-literal';
}
```

is more explicit than 

```ts
type Colors = {
  default: string;
}
```

With the former it'll be very easy in the ID to see what hex color is being used and with the latter, all you'll see for the type is "string".

vietnogigoodrx: To make sure we're on the same page, I want to make sure the original comment was suggesting:

```
const defaultColors = {
  default: '#494D55',
  black: '#1B202B',
} as const;

type ThemeColors = typeof defaultColors
```

instead of

```
interface ThemeColors {
  default: string;
  black: string;
}

const defaultColors: ThemeColors = {
  default: '#494D55',
  black: '#1B202B',
}
```

erikshestopal: @vietnogigoodrx I created a [Codesandbox](https://codesandbox.io/s/styled-components-typescript-ntpgg?file=/src/App.tsx) to demonstrate what I mean 😄

vietnogigoodrx: Cool, I took a look and added couple of lines (`const myCustomTheme: Theme = {`) to illustrate a problem.

vietnogigoodrx: @erikshestopal👆🏻

erikshestopal: @vietnogigoodrx You're gonna have fork the sandbox and post another link because I can't see your changes.

vietnogigoodrx: @erikshestopal can you see the code changes in this link?  https://codesandbox.io/s/styled-components-typescript-csdby?file=/src/App.tsx

erikshestopal: @vietnogigoodrx Here's how I would approach that problem: https://codesandbox.io/s/styled-components-typescript-46y3w?file=/src/App.tsx

vietnogigoodrx: @erikshestopal  That solution does not yield the same result.  I wanted to created another `Theme` type not a `DefaultTheme` type.

erikshestopal: I can understand overriding the default `molekule` theme, say in the EMR where we have different colors, but why would we have a second theme? My proposed solution provides the ability to override the default theme per app if we need different values.

vietnogigoodrx: If I were to override the theme, I would use the `Theme` type to tell me which keys I can override and what type it should be.  In the proposed solution, 

```
const customTheme: DefaultTheme = {
  ...theme,
  colors: {
    ...theme.colors,
    primaryLight: "white",
    primary: 123,
    randomeKey: 'foo',
  }
};
```
has no type checking.

vietnogigoodrx: @erikshestopal ☝🏻

erikshestopal: @vietnogigoodrx I'm not sure I follow because I'm getting the typechecking that you're looking for 🤔 

Screen Shot 2020-05-27 at 2 08 13 PM

vietnogigoodrx: @erikshestopal hmm, thats odd.  `DeefaultTheme` is pretty much any object so it shouldn't error.  This is whats showing up for me:

typing

kylealwyn:
Bump 🙃

vietnogigoodrx:
Was going to let @topmillerm2 take over this PR since he was free, but I can take it back. 😀

vietnogigoodrx:
@erikshestopal I made an update to use `as const` for the constants I still ended up having to define the types. What you think?  @topmillerm2 @wjossey @albertovillalobos @choochootrain Can I get some feedback for this PR?

wjossey:
I think the PR looks 👍 I'd rather someone more intimate with the code give you the final OK, but I'm a yes.

src/types/ThemeBreakpoints.ts
```diff
@@ -0,0 +1,7 @@
+import { defaultBreakpoints } from '../defaultBreakpoints';
+
+type Transform<T> = {
+  [P in keyof T]: string;
+};
+
+export interface ThemeBreakpoints extends Transform<typeof defaultBreakpoints> {}
```
erikshestopal, line 7: What's the purpose of these `Transform` generics?

vietnogigoodrx: Create a type based on the keys of defaultBreakpoints

erikshestopal: Curious why we need a separate type in addition to `defaultBreakpoints`?

vietnogigoodrx: I dont think we're able to dynamically create a theme (createTheme.ts) if we use types that are constants.

vietnogigoodrx:
bump

src/Icon/Icon.tsx
```diff
@@ -1,43 +1,56 @@
...
-        css`
-          pointer-events: none;
-          opacity: 0.65;
-        `};
+const Icon: IconType = Object.assign(
+  createComponent<IconProps, 'i'>({
+    name: 'Icon',
+    tag: 'i',
+    props: ({ name }) => ({
+      className: Icon.getClassName(name),
```
erikshestopal, line 21: Instead of the `Object.assign` workaround, can we inline the `className` generation logic? I don't think we use this static `getClassName ` function outside of this component so it seems safe to do so. 
```suggestion
      className: `mdi mdi-${name}`,
```

vietnogigoodrx: @erikshestopal  Is this true for the rest of the components?

vietnogigoodrx: @erikshestopal ignore question, doesnt seem like any other components use it 😄

vietnogigoodrx: @erikshestopal Removed getClassName, glad to see it gone.

src/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  name: string
+) => `${className || ''} ${classPrefix}-${name} ${variant ? `${classPrefix}-${name}-${variant}` : ''}`.trim();
+
+interface CreateComponentProps {
+  name: string;
+  tag?: keyof JSX.IntrinsicElements;
+  as?: React.ComponentType<any>;
+  style?: any;
+  props?: (props: any) => any;
+}
```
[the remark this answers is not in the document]
vietnogigoodrx: Suggestion looks good; however, I would like to wrap this PR up and get in merged in.  I will create another PR with these suggestions and add more typings.

erikshestopal (approved):

PR #136, opened 2020-07-27
chore: Add some :sparkles: to Typescript conversion 

erikshestopal:
Extending some of Alex's awesome work to add stricter type safety and use some different approaches to reduce boilerplate.


src/createTheme.spec.ts
```diff
@@ -16,7 +17,7 @@ describe('createTheme', () => {
         ...defaultThemeColors,
         primaryDark: 'foo',
       },
-    });
+    } as const);
 
     expect(theme.colors.primaryDark).toEqual('foo');
```
vietnogigoodrx, line 22: @erikshestopal Do you get types for `theme.colors.primaryDark` ? My VsCode is treating it as `any`.

erikshestopal: I had to restart the TS server 
Screen Shot 2020-07-27 at 12 08 20 PM

src/createTheme.spec.ts
```diff
@@ -16,7 +17,7 @@ describe('createTheme', () => {
         ...defaultThemeColors,
         primaryDark: 'foo',
```
vietnogigoodrx, line 18: @erikshestopal  I dont get an error if typed the key wrong, `primaryDark` -> `priimaryDark`

erikshestopal: Screen Shot 2020-07-27 at 12 20 24 PM

erikshestopal: Or are you saying if you add the typo to the override `colors` object?

vietnogigoodrx: I meant online 18

erikshestopal: @vietnogigoodrx do you think you can fix that issue? I hacked this over the weekend and didn't address small edge cases. I have some other priorities that I have to work on but would love to help sometime later this week.

PR #138, opened 2020-07-31
chore: Convert Components A-D to TS (WIP)

rkrishn7:
### 📃 Summary

Still a work in progress, but feel free to review these commits that I'll be rolling out.

### 🔍 Changeset

- Converted `Accordion` to TS. Made `Accordion` a functional component. In this case, since we weren't implementing any optimizations, I think it makes sense.

Note: Update `createComponent` so we can get accurate typings.

### 📈 Risk Analysis 
- [ ] Does this mutate data? 
- [ ] Does it touch PHI? 
- [ ] Does it touch payments? 


src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,133 @@
+import React, { useState, useCallback } from 'react';
```
vietnogigoodrx, line 1: You may want to check out src/Icon/Icon.tsx

rkrishn7: > You may want to check out src/Icon/Icon.tsx

Regarding specifying template types for `createComponent`? Or was there something else I'm missing?

vietnogigoodrx: Nothing in particular, just FYI for you so you can reference it.

PR #147, opened 2020-08-10
feat: auto generate prop info

rkrishn7:
### 📃 Summary

Added `react-docgen-typescript-loader` and `@storybook/addon-info` to auto generate prop tables in storybook. This works with JS implementations but plays way nicer with TS components. For a good reference, check out `Icon` as it has already been converted.

### 📸 Screenshots

Screen Shot 2020-08-10 at 3 55 10 PM


vietnogigoodrx:
Is this based on PropTypes?

vietnogigoodrx:
nm, its based on TS

rkrishn7:
> Curious how useful this is because vscode already provides typing documentation that covers the entire codebase in addition to components.

AFAIK vscode provides intellisense and definition peeking to aid with this, but personally I've found relying solely on that to be annoying. Whether it be our own properties or those inherited from `styled-system`, I think it's nice to have one centralized place for all of this.

> So in the screenshot above, how do we see documentation on CSSProperties?

Assuming you're talking about `styled-system` props, we can see that information as well.

Screen Shot 2020-08-11 at 10 00 21 AM

vietnogigoodrx:
ok, lets give this a try

rkrishn7:
I'll hold off on merging this until [this](https://github.com/heydoctor/molekule/pull/150) has been merged

PR #154, opened 2020-08-18
chore: types A-D

rkrishn7:


rkrishn7:
@vietnogigoodrx we'll need to wait for `Flex` to be converted before merging as this is the cause of most of the errors right now.

Also, do you know what this does in `src/Card/Card.tsx`?

```ts
& + ${Card.Footer} { (line 52)
```

vietnogigoodrx:
@rkrishn7 Its styled-components overloading the .toString method, the codeis equivalent to `& + ${Card.Footer.toString()} {`

vietnogigoodrx:
```
const cardFooterClass = Card.Footer.toString();

Card.Body = createComponent<BoxProps>({
  name: 'CardBody',
  as: Box,
  style: () => css`
    padding: 16px;

    & + ${cardFooterClass} {
      padding-top: 0px;
    }
  `,
});
``` should work and the errors gone

vietnogigoodrx:
@rkrishn7 I am concern with the `styles={styles}` error. Looks like the error is the result of changes in `createComponent`

vietnogigoodrx:
@rkrishn7 added a commit for flex and fix the card error.

PR #155, opened 2020-08-18
feat: add boxed checkbox styling

gerrymi:
## 📃 Summary
Added a boolean prop to enable boxed styling, along with a width prop.
boxed-checkbox

### 🏷 Asana Task
https://app.asana.com/0/1179589169762852/1188967808291211/f
	
### 📈 Risk Analysis
- [ ] Does this mutate data?
- [ ] Does it touch PHI?
- [ ] Does it touch payments?

### DIFF
--- a/src/Form/Checkbox.js
+++ b/src/Form/Checkbox.js
@@ -125,7 +125,7 @@ const CheckboxLabel = createComponent({
 const CheckboxContainer = createComponent({
   name: 'CheckboxContainer',
   tag: 'label',
-  style: ({ theme, isChecked, isDisabled, isHorizontal, size, color }) => {
+  style: ({ theme, isChecked, isDisabled, isHorizontal, size, color, isBoxed, width }) => {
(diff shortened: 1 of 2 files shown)


vietnogigoodrx:
I am thinking this checkbox design should not live in molekule because its not used for the webapp.

vietnogigoodrx:
Can we confirm with the design team that they are planning to have the same checkbox design in the webapp too?

gerrymi:
@vietnogigoodrx Waiting on design to get back to me on that, although I don't necessarily agree that only designs used in both apps warrant being included in molekule... I think if something is more than a one off that doesn't affect other functionality, it should be able to live in molekule.

vietnogigoodrx:
@gerrymi My concern is that Molekule is suppose to be the base component. This PR appears to be adding a customization to the base design which is breaking our philosophy.

gerrymi:
@vietnogigoodrx if that's the case then it doesn't matter if web-app implements it because it's not considered "base" design.

